Issue #58, opened 2015-02-02
company reference: Explain that default copany is created automatically


teemow:
By "Velrok" on gitter:

> In http://docs.giantswarm.io/reference/registry : talks about companies; and in fact you are already doing the intuitive thing of automatically creating a company with ones user name :thumbsup: . For extra clarity it would be great for that page to say that it is so. "BTW we already creates and assign you a company identical to your user name." Or something similar :smile: .

Original Author: @marian
Created at 08-12-2014 08:39:21
Gitlab: [giantswarm/docs!19](https://git.giantswarm.io/giantswarm/docs/issues/19)

Issue #38, opened 2015-02-02
Let search find gigantic


teemow:
http://docs.giantswarm.io/search/?q=gigantic/ should find domains in swarm.jsons

Possible problem / solution: configure elasticsearch tokenizer for dots.

Original Author: @luebken
Created at 10-12-2014 15:08:23
Gitlab: [giantswarm/docs!27](https://git.giantswarm.io/giantswarm/docs/issues/27)

PR #136, opened 2015-05-29
Fixing #38 and #58 (searching "foo" should also find "foo.bar")

marians:
Work in progress

Fixes https://github.com/giantswarm/docs-content/issues/58


marians:
Old and moldy, closing

PR #246, opened 2018-08-02
Several changes to formatting and structure

marians:
This should improve clarity and understandability.

In addition, it controls the syntax highlighting of some code blocks. By adding `nohighlight` to the opening backticks, we turn off highlighting, as the highlighter would otherwise try to guess a languages, which always fails. In the case of JSON, for example, we can instead enforce the language by adding `json` to the opening backticks.

For the `az role definition create --role-definition @guest.json` I removed the output, as it seems not relevant (the user doesn't have to work with it). We should mention that the output can be discarded. See the TODO.

In addition, this moves the image to the right place.


marians:
@TheoBrigitte Feel free to merge this into you branch yourself.

src/content/guides/prepare-azure-account-for-guest-clusters/index.md
```diff
@@ -34,57 +34,34 @@ To create a service principal you need:
...
-      "notDataActions": []
-    }
-  ],
-  "roleName": "azure-operator",
-  "roleType": "CustomRole",
-  "type": "Microsoft.Authorization/roleDefinitions"
-}
 ```
 
+TODO: explain what to do with the JSON output, or explain that the output can be discarded.
```
TheoBrigitte: We could mention here that `"roleName": "azure-operator"` is to be used in the next step for the role argument (`--role "azure-operator"`). So we are safe in the eventuality that we update the role name in our template file.

marians: Hmm. That name `azure-operator` is coming from our own template, isn't it? If yes, then I don't see a need to mention it here again.

TheoBrigitte: Ok

PR #248, opened 2018-08-20
Remove storage info for AWS in 'gsctl show cluster'

marians:
For https://github.com/giantswarm/giantswarm/issues/3953

### DIFF
--- a/src/content/reference/gsctl/show-cluster.md
+++ b/src/content/reference/gsctl/show-cluster.md
@@ -63,7 +63,7 @@ The output lines in detail:
 - **Worker VM size:** (_only on Azure_) VM size used for worker nodes
 - **CPU cores in workers:** total number of CPU cores in all worker nodes in this cluster
 - **RAM in worker nodes (GB):** total amount of memory in all worker nodes in this cluster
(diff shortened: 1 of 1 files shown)


src/content/reference/gsctl/show-cluster.md
```diff
@@ -64,7 +63,7 @@ The output lines in detail:
 - **Worker VM size:** (_only on Azure_) VM size used for worker nodes
 - **CPU cores in workers:** total number of CPU cores in all worker nodes in this cluster
 - **RAM in worker nodes (GB):** total amount of memory in all worker nodes in this cluster
-- **Storage in worker nodes (GB):** total amount of local storage in all worker nodes in this cluster
+- **Storage in worker nodes (GB):** (only for KVM) total amount of local storage in all worker nodes in this cluster
```
kopiczko: Above we have `**Worker VM size:** (_only on Azure_)` (not undrescores). I think would be good to take consistent approach. I personally don't case which one.

marians: Thanks, changed!

PR #253, opened 2018-08-29
GDPR banner update August 2018

marians:
For https://github.com/giantswarm/giantswarm/issues/3676

The bigger logic change here comes from switching to version 6 of privacy.js as introduced in giantswarm/web-assets#30


marians:
Here CircleCI fails, as the base Docker image `nginx:1-alpine` can't be pulled currently.

PR #262, opened 2018-09-24
Add istio docs for tenant clusters

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/4224


marians:
There is `istio` in this PR title and in the file name, but it's nowhere in the content. Are you sure you submitted the right content?

pipo02mix:
Sorry I copy from docs-content stale repo to this new one and I copy the wrong file. Fixed! PTAL

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
++++
+title = "How to deploy Istio in your cluster"
+description = "Tutorial on how to deploy Istio in the differnet providers offered by Giant Swarm."
```
marians: "in the differnet providers offered by Giant Swarm" - that doesn't sound right (and the spelling for different isn't right). I'd prefer "supported" over "offered". Also, the cloud providers don't play a big role here in my opinion. How about:

    Tutorial on how to deploy Istio on a Giant Swarm Kubernetes cluster

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+date = "2018-09-27"
+type = "page"
+weight = 30
+tags = ["tutorial"]
++++
+
+# Preparation
```
marians: The first H1 headline (one `'#`) should be the title of the article. If "Preparation" is the first content section of the article, it should be a H2 (`##`) after the article headline.

marians: In terms of content, I would like to greet readers who don't know a lot about Istio with one or two general sentences on what it is, why you would want to look into it and why you need special installation instructions for Giant Swarm.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+Nowadays, the recommended way to deploy Istio is using Helm. Please ensure the Kubernetes package manager is installed and running in your cluster. Go to the official [docs](https://docs.helm.sh/using_helm#install-helm) in case you need to install it.
+
+Now, it is time to [download the Istio project package](https://github.com/istio/istio/releases/) and include the command line interface in your path. After having it in your local uncompressed, enter in the project folder and make sure the `/bin` directory is under your system path.
```
marians: > After having it in your local uncompressed

Missing word?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+Verify you are ready running these commands
```
marians: Please add a blank line before and after every code block for better readability

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+# Check helm client and tiller are running
```
marians: These comments in the beginning of a code block should not be necessary. The normal text should explain what the command is doing. Here, in this case, the comment more or less duplicates the sentence before.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+$ helm version
+Client: &version.Version{SemVer:"v2.10.0", GitCommit:"X", GitTreeState:"clean"}
+Server: &version.Version{SemVer:"v2.10.0", GitCommit:"X", GitTreeState:"clean"}
+
+# Check istioctl is available
+$ istioctl version
+Version: X.X.X
+...
+BuildStatus: Clean
+```
```
marians: I would separate the `istioctl version` part into its own code block vor better structure. Again, the comment content belongs in the normal article text bertween the code blocks.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+  name: istio
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: <NAMESPACE_OF_YOUR_SERVICES>
+```
+
+## Recomendations 
+
+The deployment is made up of a different number of components. Some of them, as pilot, have a large impact in terms of memory and CPU, so it is recommended to have around 8GB and 4CPUs free in your cluster. Obviously, all components have `requested resources` defined, so in case you don't have enough capacity you will see pods not starting.
```
marians: `8GB and 4CPUs` -> `8 GB of memory and 4 CPUs`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+A followed convention, not only within Istio community but Kubernetes too, is to define in all your app deployments the label `app` and `version` as metadata. Both labels help the tracing and metric system to have richer meta information. 
+
+Another good practice is to name the service ports. Istio uses the name to discover the protocol used by the end service container. Thus you have to prefix the port name with the protocol desired. The different supported protocols (http, http2, grpc, mongo, or redis) leverage Istio to route traffic more intelligently. Otherwise, it will use plain TCP routing.
```
marians: Does istio expect certain protocol names here? If yes, we should make that more explicit, format the example names as literals (in backticks) and probably link to a full reference.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+  name: app
+spec:
+  ports:
+  - name: http-app
+    port: 80
+    protocol: TCP
+    targetPort: 80
+```
+
+In the strange case, you use multiple services laying over the same pods, you have to use the same kind of traffic for the ports defined in the service.
```
marians: `In the strange case, you use multiple services laying over the same pods` - I would rephrase that to `If you have multiple services exposing port of the same pods`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+metadata:                  metadata:
+  name: app-svc-1            name: app-svc-2
+spec:                      spec:
+  ports:                     ports:
+  - name: http-app           - name: http2-app
+    port: 80                   port: 80
+    protocol: TCP              protocol: TCP
+  selector:                  selector:    
+    app: my-app                app: my-app
+```
```
marians: This code block doesn't make sense to me. Is one right and one wrong? Why are the next to each other?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+Finally, it is good to mention how you can integrate Istio within your existing workloads. By default mutual TLS is not configured, mainly for avoiding downtime issues it can provoke. But lately, there has appeared a new option which you can set the TLS policy as `permissive` and let the proxy accept traffic encrypted and unencrypted. Thus safe wise you can start with this policy and slowly move your services to a `strict` mode TLS in case it is required.
```
marians: `Finally, it is good to mention how you can integrate` -> `Finally you may be interested in how you can integrate`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+- Mixer
+- Pilot
+- Grafana
+- Prometheus
+- Tracing (Jaeger) 
+- Kiali
+- Mutual TLS
+- Certmanager
+
+All of those are deployed but mutual TLS and certmanager. You can configure the components tweaking the values passed to helm. Run `helm inspect install/kubernetes/helm/istio` to display the different options available.
```
marians: `All of those are deployed but mutual TLS and certmanager.` -> `With exception of Mutual TLS and Certmanager, all of those are installed.`

Question: What's with the two exceptions? Can they be installed manually? Is it recommendable?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+  name: istio
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: istio
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: <NAMESPACE_OF_YOUR_SERVICES>
+```
```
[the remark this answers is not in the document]
puja108: he explained it above, but it's not clear that this is what he explains

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+- Grafana
+- Prometheus
+- Tracing (Jaeger) 
+- Kiali
+- Mutual TLS
+- Certmanager
+
+All of those are deployed but mutual TLS and certmanager. You can configure the components tweaking the values passed to helm. Run `helm inspect install/kubernetes/helm/istio` to display the different options available.
+
+Although Istio can run in almost every Kubernetes cluster no matter the provider, there are some quirks to consider depending on where the service mesh is deployed.
```
marians: `no matter the provider` -> `regardless of the underlying infrastructure`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+### AWS and Azure
+
+```
+$ helm install install/kubernetes/helm/istio --name istio --namespace istio-system
+```
```
marians: This is missing some explanation. What does this do that is specific to AWS and Azure?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Giantnetes
```
marians: If this is the KVM-related section, I think the title should be `KVM` or `Base metal (KVM)` or `On premises (KVM)`.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+```
+$ helm install install/kubernetes/helm/istio --name istio --namespace istio-system \
+--set gateways.istio-ingressgateway.type=NodePort --set gateways.istio-egressgateway.type=NodePort 
+```
```
marians: Again, this needs some explanation

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+Ensure all components are up and running listing the pods in the istio namespace.
+```
+$ kubectl get pod -n istio-system -w
```
marians: Even if `kubectl get pod` would work, I think `kubectl get pod` is more explicit in communication that this is about listing multiple pods.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+## Deploy an application
+
+Let's verify Istio is deployed and configure correctly. We can deploy a simple http server that returns `200` code for the `healthz` path of the service.
+
+### Automated sidecar injection
+
+First of all we need to label the namespace to make easier the sidecar injection.
```
marians: Is this really about making something easier, or is it require to enable the injection?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+kubectl label namespace default istio-injection=enabled
+```
+And now we can install the chart containing the example app
```
marians: Please append `.` or `:`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+$ helm registry install quay.io/giantswarm/liveness-chart
+```
+
+### Manual sidecar injection
+
+In case you want the manual injection, pull the chart and use istio command line to parse the deployment and insert the proxy.
```
marians: `use istio command line` -> `use the Istio command line utility`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+# (GIANTNETES) Parse the chart template to Kubernetes resources (use the ingress port given by support)
+$ helm template /tmp/liveness/giantswarm_liveness-chart_XXX/liveness-chart /tmp/liveness.yaml --set ingress.port=30010
+
+# Inject the sidecar in the deployment and submit the resources to Kubernetes API
+$ istioctl kube-inject -f /tmp/liveness.yaml | k apply -f -
+```
+
+It will launch a deployment and service listening in the port 80. At the same time, it will create a virtual service redirecting the traffic to the liveness service. Also, there is a gateway to wire the virtual service up with the ingress gateway.
+
```
marians: `listening in the port 80` -> `listening on port 80`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+### AWS verification
+
+Get service hostname given by the cloud provider. Run
```
marians: `Get service hostname given by the cloud provider` -> `Get the load balancer hostname`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+$ kubectl get svc istio-ingressgateway -n istio-system -o jsonpath="{.status.loadBalancer.ingress[0].hostname}"
```
marians: `kubectl get svc` -> `kubectl get service` (use explicit standard naming in favour of abbreviation)

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+And it will return the URL which the deployed app should reply. As we have set wildcard `*` in the hostname of the virtual service all `/healthz` traffic will be forwarded to the service.
```
marians: `And it will return the URL which` -> `This will return the URL under which`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: istio
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: <NAMESPACE_OF_YOUR_SERVICES>
+```
+
+## Recomendations
```
puja108: `Recomendations` -> `Recommendations`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+And it will return the URL which the deployed app should reply. As we have set wildcard `*` in the hostname of the virtual service all `/healthz` traffic will be forwarded to the service.
+
+Executing this curl command should give you a `200` response.
+```
+curl http://<ELB_HOSTNAME>/healthz -I
+```
+
+### Azure verification
+
+Get the ingress controller IP given by the cloud provider. Run
```
marians: `given by the cloud provider` -> `assigned by Azure`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+$ kubectl get svc istio-ingressgateway -n istio-system -o jsonpath="{.status.loadBalancer.ingress[0].ip}"
+```
+And it will return the URL which the deployed app should reply to. Same as AWS, the wildcard `*` set as the hostname in the virtual service will send all traffic from `/healthz` path to the liveness service.
+
+Running this curl command should give you a `200` response.
+```
+curl http://<IP>/healthz -I
+```
+
+## On premise (Giantnetes)
```
marians: Re: headline naming, see above

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+As probably you may know our [`giantnetes`](https://docs.giantswarm.io/basics/onprem-architecture/) tenant pods live inside the control plane pods. It means Giant Swarm creates a forwarding between the external ingress service to the ingress placed inside the tenant cluster. By default, the tenant clusters has a nginx ingress controller allocated out of the box. But, in case you want to use istio ingress controller you need to change the current ingress controller (remember it will affect all current traffic for that tenant cluster) or ask our team to allocate a new redirection from the parent endpoint to the Istio controller.
```
marians: - Again, please replace `giantnetes` with whatever you choose above
- `As probably you may know our` -> `As you may probably know, our`
- `has a nginx ingress` -> `has an nginx ingress`
- `use istio ingress controller` -> `use the Istio ingress controller`
- Please break the last sentence apart in several to improve readability

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+
+Modifying the port for the ingress gateway
+```
+kubectl edit svc istio-ingressgateway -n istio-system
```
marians: `kubectl edit svc` -> `$ kubectl edit service`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+    nodePort: <PORT_PROVIDED_BY_GS_OR_DEFAULT_ONE>
+    port: 80
+    protocol: TCP
+    targetPort: 80
+  - name: https
+    nodePort: <PORT_PROVIDED_BY_GS_OR_DEFAULT_ONE>
+    port: 443
+    protocol: TCP
+    targetPort: 443
+----
```
marians: Since the manifest snippet here is not terminal output, I would move it into it's own code block and add a sentence of explanation in top.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+If you run this curl command you should get a `200` response.
+```
+curl http://<LOAD_BALANCER_IP>:<PORT_TENANT_CLUSTER>/healthz -I
+```
+
+Giant Swarm encourages you to leave this liveness app up and running in order to make our monitoring happy. By default, nginx ingress controller provides a `/healthz` endpoint to check ICs are up, but Istio controller lack of this functionality. For that reason, the liveness app will give us the health signal required to have the correct observability of your istio clusters.
```
marians: `but Istio controller lack of this` -> `but the Istio ingress controller lacks this`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+## Troubleshooting
+
+- The pods cannot start with injected sidecars. Looking in the deployment status you see
+```
+message: 'pods "details-v1-7986ddbd99-" is forbidden: unable to validate against
+   any pod security policy: [spec.initContainers[0].securityContext.privileged:
+   Invalid value: true: Privileged containers are not allowed capabilities.add:
+   Invalid value: "NET_ADMIN": capability may not be added spec.initContainers[1].securityContext.privileged:
+```
+It means the pod security policy is not bind to the service account used by the deployment. Please review the `Preparation` section.
```
marians: `not bind to` -> `not bound to`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+- The request are not routed to the ingress gateway. In some cases, the default gateway is not proper configured. If istio has just been deployed, try to delete it and check now the status.
```
marians: `The request are not routed` -> `Requests are not routed`

marians: `proper configured` -> `configured properly`

marians: `check now the status` -> `check the status again`

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+```
+$ kubectl delete gateways.networking.istio.io istio-autogenerated-k8s-ingress -n istio-system
+```
```
marians: We need some explanation above this block. Or is it related to the text above? Then please change that sentence to something like `check the status again using the command below:`

marians:
I'm done with my first review. Lots of small comments, nothing really big. I can see this becoming a very good resource for our customers!

Once you are done with changes, feel free to re-ping. I'll give the final formatting and code highlighting another check then.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+Server: &version.Version{SemVer:"v2.10.0", GitCommit:"X", GitTreeState:"clean"}
+
+# Check istioctl is available
+$ istioctl version
+Version: X.X.X
+...
+BuildStatus: Clean
+```
+
+Secondly, in order to avoid a known issue with the sidecar security context (privileges), it is needed to extend or create the pod security policy tha will be assigned to your micro services. Istio, beyond the sidecar container, injects a init container in the app's deployment resource. It takes care of forwarding traffic between the proxy and the original container(s).
```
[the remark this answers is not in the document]
pipo02mix: But container can be one or more

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+
+Another good practice is to name the service ports. Istio uses the name to discover the protocol used by the end service container. Thus you have to prefix the port name with the protocol desired. The different supported protocols (http, http2, grpc, mongo, or redis) leverage Istio to route traffic more intelligently. Otherwise, it will use plain TCP routing.
+```
+apiVersion: v1
+kind: Service
+metadata:
+  name: app
+spec:
+  ports:
+  - name: http-app
```
puja108: let's do

`name: app` -> `name: myapp`

`http-app` -> `http-myapp`

this way it's more clear

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+    port: 80
+    protocol: TCP
```
puja108: this does not make sense, above you say http, now you say TCP

src/content/guides/deploying-istio/index.md
puja108: why would you ever have 2 services for the same port? what would be a use case?

src/content/guides/deploying-istio/index.md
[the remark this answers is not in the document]
pipo02mix: I wrote `Then, define an istio role and a role binding for wiring it together against the default service account for the namespace where your apps are running. `

What you will modify to emphasize it has to be replaced?

src/content/guides/deploying-istio/index.md
puja108: `a new option which you can set` -> `a new option where you can set`

`accept traffic encrypted and unencrypted` -> `accept both encrypted and unencrypted traffic`

`Thus safe wise` -> `Thus, to be on the safe side,`

src/content/guides/deploying-istio/index.md
pipo02mix: yeah I wanted to add an example of the wrong configuration to make clear my point. Here, I exposed two services pointing same app (selector matches) but using different protocols

pipo02mix: It took time to me for understanding (because Istio docs are short in this matter) and I wanted to give more clues 

Ideas welcome

pipo02mix: I showed them in columns to make clearer the difference. But it looks it does not work 😓

src/content/guides/deploying-istio/index.md
pipo02mix: Mutual TLS is explained in recommendations. Cert manager is like an addon. I can add more info .good point

src/content/guides/deploying-istio/index.md
pipo02mix: I think you meant `pods`, right?

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,283 @@
...
+    image: docker.io/istio/proxy_init:0.6.0
+    name: istio-init
+    securityContext:
+      capabilities:
+        add:
+        - NET_ADMIN
+      privileged: true
+```
+
+As consequence, the pod now must have the right security policies configured to run in the cluster. An easy way to solve the problem is to create a pod security policy that contains the correct capabilities. Then, define an istio role and a role binding for wiring it together against the default service account for the namespace where your apps are running.
```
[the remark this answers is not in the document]
pipo02mix: But the psp example is below

src/content/guides/deploying-istio/index.md
pipo02mix: well one is k8s level and the other istio level

pipo02mix: but I will omit as it is the default

src/content/guides/deploying-istio/index.md
pipo02mix: I have really no idea. I have just added it because the case is mentioned in istio docs 

[Service association](https://istio.io/docs/setup/kubernetes/spec-requirements/)

pipo02mix: I am not sure if I got it right, but I think it is not a usual practice. We can omit it too. WDYT?

pipo02mix:
thx for the feedback!
I have worked in the corrections and I have done another scan. PTAL

marians:
Some more general comments:

- The `weight` attribute in the frontmatter should be higher. The current value places this page right in the top of all articles on the `/guides/` page. I think the content is not of such a broad interest, compared to some of the other guides. Please adjust to place the article somewhere in the lower end.

- The syntax highlighting of code blocks currently looks odd, with the exception of the yaml blocks. Let's add use the `nohighlight` identifier in the beginning of code blocks to turn it off.

- Terminal commands should start with `$`.

- Let's add the typical `Further reading` section at the bottom with a few helpful links. See [here](https://docs.giantswarm.io/guides/prepare-aws-account-for-tenant-clusters/#further-reading) for an example.

- Please adapt the front matter date when actually publishing.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Preparation
+
+Nowadays, the recommended way to deploy Istio is using Helm. Please ensure the Kubernetes package manager is installed and running in your cluster. Go to the official [docs](https://docs.helm.sh/using_helm#install-helm) in case you need to install it.
+
+Now, it is time to [download the Istio project package](https://github.com/istio/istio/releases/) and include the command line interface in your path. After having it in a local folder uncompressed, make sure the `/bin` directory is under your system path.
+
+Let's verify the environment is ready. First check helm is installed correctly.
+
+```
```
marians: Here, for example, please add `nohighlight` to the opening backticks.

    ```nohighlight

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,312 @@
...
+$ helm version
+Client: &version.Version{SemVer:"v2.10.0", GitCommit:"X", GitTreeState:"clean"}
+Server: &version.Version{SemVer:"v2.10.0", GitCommit:"X", GitTreeState:"clean"}
+```
+
+Now, ensure Istio `cli` is in your path.
```
marians: > ensure Istio `cli` is

ensure `istiocli` is

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,312 @@
...
+
+# (CLOUD) Parse the chart template to Kubernetes resources
+$ helm template /tmp/liveness/giantswarm_liveness-chart_XXX/liveness-chart /tmp/liveness.yaml
+
+# (GIANTNETES) Parse the chart template to Kubernetes resources (use the ingress port given by support)
+$ helm template /tmp/liveness/giantswarm_liveness-chart_XXX/liveness-chart /tmp/liveness.yaml --set ingress.port=30010
+
+# Inject the sidecar in the deployment and submit the resources to Kubernetes API
+$ istioctl kube-inject -f /tmp/liveness.yaml | k apply -f -
+```
```
marians: I still think what's in the comments here should just be normal article text.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,312 @@
...
+description = "Tutorial on how to deploy Istio on a Giant Swarm Kubernetes cluster."
+date = "2018-09-27"
+type = "page"
+weight = 30
+tags = ["tutorial"]
++++
+
+# How to deploy Istio in your cluster
+
+After microservices has conquered the world, another dimension of concerns has arisen. The technology has allowed us to build distributed architectures thanks to tools like Kubernetes. Although the orchestrator helps us with the deployment and maintenance of our services there a bunch of other concerns we have to deal. As an example, most people use an insecure protocol to communicate their services. For that and many other reasons, the service meshes have appeared with the goal of helping to manage these distributed scenarios.
```
marians: `After microservices has` -> `After microservices have`

marians: `there a bunch of other concerns` -> `there are a bunch of other concerns`

marians: `we have to deal.` -> `we have to deal with.`

src/content/guides/deploying-istio/index.md
pipo02mix: forgot those ones

pipo02mix:
Second round done

src/content/guides/deploying-istio/index.md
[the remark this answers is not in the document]
marians: I guess it's how it would be understood, but it just makes it harder to read. My thinking is if you just use plural, everybody will understand that the traffic can also come from a single container in a special case.

src/content/guides/deploying-istio/index.md
marians: How likely is this to happen BTW? Should this just be an item in the Troubleshooting section? And I think the example is not needed here, as it's not illustrating a lot.

pipo02mix: ok I like the idea of moving it to the Troubleshooting

pipo02mix: done

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,332 @@
...
+...
+```
+
+After configured it, running the next command you should get a `200` response.
+
+```nohighlight
+$ curl http://<LOAD_BALANCER_IP>:<PORT_TENANT_CLUSTER>/healthz -I
+```
+
+Giant Swarm encourages you to leave this liveness app up and running in order to make our monitoring happy. By default, nginx ingress controller provides a `/healthz` endpoint to check ICs are up, but the Istio controller lacks of this functionality. For that reason, the liveness app will give us the health signal required to have the correct observability of your istio clusters.
```
puja108: I guess this is obsolete if we do not support replacement of the IC for now. We can think of adding this once we can actually make IC optional

tommyhobin:
I'll take a look and make changes for review.

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,332 @@
...
+
+The actual installation is quite simple thanks to helm. The following command will deploy all Istio components in a new namespace reserved.
+
+```nohighlight
+$ helm install install/kubernetes/helm/istio --name istio --namespace istio-system
+```
+
+### On premises (KVM)
+
+The main difference with clouds is the ingress gateway service must be type `NodePort`. Since there is no automatic mechainism to provide an endpoint, the service is exposed in the host underlaying machine. Later, it is explained how to redirect the traffic from the control plane to a second tenant cluster ingress controller.
```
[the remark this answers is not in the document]
pipo02mix: I wanted to highlight that clouds automatically provide you an endpoint (via the load balancer) but on KVM the traffic is routed from the control plane machines (VMs) to the inner tenant machine (VM too hehe). I changed the machine by VM, is it clearer? how do you write it up?

pipo02mix:
@puja108 PTAL

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,334 @@
...
+After obtaining the ports, modify the ingress gateway to set the correct configuration.
+
+```nohighlight
+$ kubectl edit service istio-ingressgateway -n istio-system
+```
+
+```yaml
+...
+  - name: http2
+    nodePort: <PORT_PROVIDED_BY_GS_OR_DEFAULT_ONE>
```
puja108: rm _OR_DEFAULT_ONE here and below

src/content/guides/deploying-istio/index.md
```diff
@@ -0,0 +1,334 @@
...
+...
+```
+
+After configuring it, running the next command should get you a `200` response.
+
+```nohighlight
+$ curl http://<LOAD_BALANCER_IP>:<PORT_TENANT_CLUSTER>/healthz -I
+```
+
+Giant Swarm encourages you to leave this liveness app up-and-running in order to make our monitoring happy. By default, nginx ingress controller provides a `/healthz` endpoint to check ICs are up, but the Istio controller lacks of this functionality. For that reason, the liveness app will give us the health signal required to have the correct observability of your istio clusters.
```
puja108: I think we can remove this now that we do not allow to replace IC

src/content/guides/deploying-istio/index.md
pipo02mix: good catch. Fixed

pipo02mix:
I added a conclusion that I think is valuable

src/content/guides/deploying-istio/index.md
```diff
@@ -289,7 +289,9 @@ After configuring it, running the next command should get you a `200` response.
 $ curl http://<LOAD_BALANCER_IP>:<PORT_TENANT_CLUSTER>/healthz -I
 ```
 
-Giant Swarm encourages you to leave this liveness app up-and-running in order to make our monitoring happy. By default, nginx ingress controller provides a `/healthz` endpoint to check ICs are up, but the Istio controller lacks of this functionality. For that reason, the liveness app will give us the health signal required to have the correct observability of your istio clusters.
+---
+
+Istio is a great project and it could mark the future of micro-service architectures, but on the other hand, it adds another layer of complexity. Giant Swarm encourages to read and understand well how it works before deploying it in your clusters.
```
puja108: maybe we should also add a disclaimer like following:

```
__Disclaimer:__ While we do not discourage installing and using istio and even document it here, we currently do not support it managed by us. That means that if you run into issues with or because of istio, this lies in your responsibility. Furthermore, as the complexity increases with istio, debugging and fixing even not directly related issue will take longer.
```

pipo02mix: nice thanks

pipo02mix:
Updated PTAL

marians:
Checking again

marians:
Just pushed a commit that changes the formatting in the troubleshooting section a bit. The bullet list doesn't work so well when it's interrupted by code blocks and longer content.

pipo02mix:
:) merging

PR #269, opened 2018-10-11
Remove font weight 100 from Roboto

marians:
For https://github.com/giantswarm/giantswarm/issues/4276


marians:
@oponder This is the last one :)

PR #275, opened 2018-10-26
Update 'gsctl show cluster' reference page

marians:
For https://github.com/giantswarm/giantswarm/issues/3987

This adds info which has been added in https://github.com/giantswarm/gsctl/pull/308


marians:
Yum!

PR #276, opened 2018-10-30
add docs for coredns configuration

puja108:
last step towards https://github.com/giantswarm/giantswarm/issues/4452

### DIFF
--- a/src/content/guides/advanced-coredns-configuration/index.md
+++ b/src/content/guides/advanced-coredns-configuration/index.md
@@ -1,0 +1,73 @@
++++
+title = "Advanced CoreDNS Configuration"
+description = "Here we describe how you can customize the configuration of the managed CoreDNS service in your clusters"
+date = "2018-10-30"
+type = "page"
+weight = 50
+tags = ["tutorial"]
++++
+
+# Advanced CoreDNS Configuration
+
+The [CoreDNS addon](https://github.com/coredns/coredns) running inside your cluster has additional configuration options and features that can be customized.
+
+You can customize two of these configuration options on a per cluster basis through a ConfigMap inside your clusters. The ConfigMap is named `coredns-user-values` and is located in the `kube-system` namespace.
+
+__Note:__ This feature is only available in more recent cluster versions. To check if your cluster version supports customization through the ConfigMap, you can check if the above-mentioned ConfigMap is present.
+
+__Warning:__ Please do not edit any of the other CoreDNS related resources. Only the user ConfigMap is safe to edit.
+
+```nohighlight
+$ kubectl -n kube-system get cm coredns-user-values
+NAME                                   DATA      AGE
+coredns-user-values                    0         11m
+```
+
+On cluster creation the ConfigMap is empty and below-mentioned defaults will be applied to the final CoreDNS deployment. To customize any of the configuration options, you just need to add the respective line(s) in the data field of the user ConfigMap.
+
+## Cache Settings
+
+By default we set the cache TTL for CoreDNS to 30 seconds. You can customize the cache settings of CoreDNS by setting the value of the cache field in the user ConfigMap like following.
+
+```yaml
+data:
+  cache: 60
+```
+
+Above setting increases the TTL to 60 seconds.
+
+The cache plugin also supports much more detailed configuration which is documented in the [upstream documentation](https://coredns.io/plugins/cache/).
+
+## Additional Proxies
+
+The default proxy entry we set in CoreDNS is
+
+```yaml
+proxy . /etc/resolv.conf
+```
+
+You can add additional proxy entries by adding a proxy list to the user ConfigMap and creating a list item for each proxy entry you want to be added.
+
+```yaml
+data:
+  proxy:
+    - foo.com 1.1.1.1
+    - bar.com 8.8.8.8
+```
+
+Above example would result in following additional proxy entries in the CoreDNS configuration:
+
+```yaml
+proxy foo.com 1.1.1.1
+proxy bar.com 8.8.8.8
+```
+
+Above setting would proxy all requests within foo.com to 1.1.1.1 which is Cloudflare's DNS and all requests within bar.com to 8.8.8.8 which is Google Public DNS. All other requests will be resolved by the default DNS provider set for your cluster.
+
+The proxy plugin also supports much more detailed configuration which is documented in the [upstream documentation](https://coredns.io/plugins/proxy/).
+
+## Further reading
+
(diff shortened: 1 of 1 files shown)


src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -0,0 +1,73 @@
...
+type = "page"
+weight = 50
+tags = ["tutorial"]
++++
+
+# Advanced CoreDNS Configuration
+
+The [CoreDNS addon](https://github.com/coredns/coredns) running inside your cluster has additional configuration options and features that can be customized.
+
+You can customize two of these configuration options on a per cluster basis through a ConfigMap inside your clusters. The ConfigMap is named `coredns-user-values` and is located in the `kube-system` namespace.
```
puja108: @TheoBrigitte is the name and namespace right?

src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+On cluster creation the ConfigMap is empty and below-mentioned defaults will be applied to the final CoreDNS deployment. To customize any of the configuration options, you just need to add the respective line(s) in the data field of the user ConfigMap.
+
+## Cache Settings
+
+By default we set the cache TTL for CoreDNS to 30 seconds. You can customize the cache settings of CoreDNS by setting the value of the cache field in the user ConfigMap like following.
+
+```yaml
+data:
+  cache: 60
```
puja108: @TheoBrigitte please check

src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -0,0 +1,73 @@
...
+proxy . /etc/resolv.conf
+```
+
+You can add additional proxy entries by adding a proxy list to the user ConfigMap and creating a list item for each proxy entry you want to be added.
+
+```yaml
+data:
+  proxy:
+    - foo.com 1.1.1.1
+    - bar.com 8.8.8.8
```
puja108: @TheoBrigitte please also check

src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -0,0 +1,73 @@
...
+__Warning:__ Please do not edit any of the other CoreDNS related resources. Only the user ConfigMap is safe to edit.
+
+```nohighlight
+$ kubectl -n kube-system get cm coredns-user-values
+NAME                                   DATA      AGE
+coredns-user-values                    0         11m
+```
+
+On cluster creation the ConfigMap is empty and below-mentioned defaults will be applied to the final CoreDNS deployment. To customize any of the configuration options, you just need to add the respective line(s) in the data field of the user ConfigMap.
+
```
pipo02mix: Dore coreDNS read the configuration automatically and does a hot reload? maybe it is worth to mention

src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -0,0 +1,73 @@
...
+```yaml
+proxy foo.com 1.1.1.1
+proxy bar.com 8.8.8.8
+```
+
+Above setting would proxy all requests within foo.com to 1.1.1.1 which is Cloudflare's DNS and all requests within bar.com to 8.8.8.8 which is Google Public DNS. All other requests will be resolved by the default DNS provider set for your cluster.
+
+The proxy plugin also supports much more detailed configuration which is documented in the [upstream documentation](https://coredns.io/plugins/proxy/).
+
```
pipo02mix: What about adding `autopath` example?

src/content/guides/advanced-coredns-configuration/index.md
puja108: chart-operator does the config update, and coredns then hot reloads AFAIK, but not sure we need to overload the technical details here

src/content/guides/advanced-coredns-configuration/index.md
puja108: what is the `autopath` example?

src/content/guides/advanced-coredns-configuration/index.md
pipo02mix: well my concern is if the user needs to restart coredns

src/content/guides/advanced-coredns-configuration/index.md
pipo02mix: https://github.com/coredns/coredns/tree/master/plugin/autopath

puja108: we only support cache and proxy config, nothing else, or do you have a customer asking for that?

pipo02mix: yes, there is one and IMO it is a nice improvement (performance wise) for external calls. More info https://github.com/giantswarm/giantswarm/issues/4540

puja108: ok, this would mean we need to add that in the code, so I would not do it now, happy to take a new issue (incl. CoreDNS update) into our board for that, and then add docs for it, but as is, we cannot support it ootb

puja108: also, best ping me on such things early on, so I have them on my radar

src/content/guides/advanced-coredns-configuration/index.md
TheoBrigitte: user does not need to restart coredns

src/content/guides/advanced-coredns-configuration/index.md
TheoBrigitte: correct :heavy_check_mark:

src/content/guides/advanced-coredns-configuration/index.md
TheoBrigitte: correct :heavy_check_mark:

src/content/guides/advanced-coredns-configuration/index.md
TheoBrigitte: wrong :negative_squared_cross_mark: 

configmap values are always string, so we can have:
* one value
```
  proxy: foo.com 1.1.1.1
```

* one value per line by using multi-line value
```
  proxy: |
    foo.com 1.1.1.1
    bar.com 8.8.8.8
```

puja108: ah, ok, thanks, wasn't sure

puja108:
@TheoBrigitte changed according to your comment, PTAL

puja108:
@marians thanks, the warning needs to move below, but the note is actually the context of the codeblock, so it needs to stay together

PR #281, opened 2018-11-22
Multiple availability zones

teemow:
Give more information about how to create and run clusters in multiple availability zones.

### DIFF
--- a/src/content/basics/multiaz/index.md
+++ b/src/content/basics/multiaz/index.md
@@ -35,6 +35,8 @@ This enables use cases such as
 
 - Availability zones get selected randomly by the Giant Swarm control plane. You only need to specify the number of availability zones. 
 
+- Nodes will get distributed evenly across availability zones. There is currently no way to determine which or how many nodes should be started in a particular availability zone. But the nodes will have a label `failure-domain.beta.kubernetes.io/zone` that indicates in which availability zone the node is running.
+
 - Single availability zone clusters start in a random availability zone too. This way we make sure that once one availability zone is unavailable not all single clusters will disappear.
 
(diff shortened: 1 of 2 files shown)


src/content/basics/multiaz/index.md
```diff
@@ -0,0 +1,56 @@
...
+weight = 50
+type = "page"
+categories = ["basics"]
++++
+
+# Clusters Over Multiple Availability Zones
+
+*This is currently only available with Giant Swarm 6.1.0 on AWS*
+
+With Giant Swarm you can easily launch cluster across multiple availability zones (AZs). This will lower the risk that your cluster will get unavailable while there is an incident with your cloud providers or your datacenters infrastructure.
```
marians: Suggesting to make it more concrete here:

```suggestion
With Giant Swarm you can easily launch cluster across multiple availability zones (AZs). This will lower the risk that your cluster will get unavailable due to an incident in a particular AWS data center.
```

src/content/basics/multiaz/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## What availability zones are good for {#benefits}
+
+Both on AWS and Azure, support AZs within their regions. These zones still have good interconnectivity but are separated from each other to isolate failures within one zone from the others. So it is more likely that one AZ goes down than the whole region. If your clusters are running within multiple AZs then Kubernetes can easily shift your workloads from one AZ into the other.
+
```
pipo02mix: Shall we mention Azure if it is not yet implemented?

src/content/basics/multiaz/index.md
pipo02mix: I would also mention that node are distributed evenly through the AZs and you can not choose where a node will be allocated

src/content/basics/multiaz/index.md
teemow: Yes I'd keep it. The interface (gsctl/happa) will contain this for Azure too. It is just that the current release only supports one AZ.

PR #292, opened 2019-01-17
Docs changes for autoscaling

marians:
For https://github.com/giantswarm/giantswarm/pull/2206

This PR intends to contain all the docs changes we need for autoscaling.

Note: This is intended to be merged into `master`, not `feature-autoscaling`.

### Changes summary

- Add page on scaling and autoscaling (`/basics/cluster-size-autoscaling/`)
- Update cluster definition YAML reference
- Update `gsctl create cluster` reference
- Update `gsctl show cluster` reference
- Update `gsctl scale cluster` reference
- Align related links format


teemow:
Nice. Thx!

I was wondering if we should emphasize or make clear that each and every cluster can have autoscaling or actually does. So once a cluster is upgraded to the AWS version that supports it you can change max and it starts autoscaling. If you don't want it anymore just turn it off by setting min==max. 

And something is wrong with the build.

marians:
> I was wondering if we should emphasize or make clear that each and every cluster can have autoscaling or actually does.

Yes. I'll add that to the first paragraph.

> And something is wrong with the build.

Should be the Github rate limiting problem. See https://gigantic.slack.com/archives/C3C7ZQXC1/p1547720912011700

src/content/reference/gsctl/create-cluster.md
```diff
@@ -27,40 +27,43 @@ The shortest possible way to create a new cluster with default settings:
...
-To set a friendly name for your new cluster, pass the `--name` argument with the `create cluster` command:
+To set a friendly name for your new cluster, pass the `--name` (or `-n`) argument with the `create cluster` command:
 
 ```nohighlight
-$ gsctl create cluster --owner=myorg --name="Test Cluster"
+$ gsctl create cluster -o myorg --name "Test Cluster"
 ```
 
-To create a new cluster with a specific number of nodes, use the `--num-workers` argument:
+To create a new cluster that auto-scales within a certain range, use the `--num-workers` and `--workers-max` argument:
```
tuommaki:
```suggestion
To create a new cluster that auto-scales within a certain range, use the `--workers-min` and `--workers-max` argument:
```
Is this the correct parameter here?

marians:
@MarcelMue @teemow Commit https://github.com/giantswarm/docs/pull/292/commits/d106bcf6adf5d8a8c40eab971169669b1df13658 adds quite some verbose info regarding the down-scaling and utilization threshold topic.

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,42 @@
...
+## Setting scaling limits
+
+With the autoscaler taking control over the number of worker nodes, you only define the bounds or limits in which the cluster size can vary. This is possible both during cluster creation as well as after creation.
+
+To enforce an exact cluster size and **effectively disable the autoscaler**, simply set the minimum and maximum worker node count to the same value. This is also the recommended way in case you want to control the number of worker nodes through some external tooling via the Giant Swarm API.
+
+## Minimal and default cluster size
+
+When creating a cluster without specifying the number of worker nodes, **three** worker nodes will be created. This is also the minimal number supported by Giant Swarm. On AWS starting with release version 6.3.0 (TODO), when not specified, the maximum number of worker nodes is also set to **three**.
+
```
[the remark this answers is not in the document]
marians: While you typed this, I changed the text accordingly after Marcel briefed me. Please check my last commit whether that's what you think we need.

And I don't exactly understand the third bullet.

src/content/reference/gsctl/create-cluster.md
marians: Thanks!

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,42 @@
...
++++
+
+# Cluster Size and Autoscaling
+
+Starting with release version 6.3.0 (TODO) for AWS, you can leverage the benefits of the [Kubernetes autoscaling components](https://github.com/kubernetes/autoscaler) to define the number of worker nodes in a cluster based on demand.
+The autoscaler is provided with every cluster of version >= 6.3.0 (TODO) on AWS.
+
+On Giant Swarm installations on Azure, on bare-metal, and on AWS prior to version 6.3.0 (TODO), the cluster size would be defined statically.
+
+**Note:** the number of master nodes cannot be changed
```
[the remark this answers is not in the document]
marians: I looked that up a while back. lowercase seems to be the rule.

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,42 @@
...
+description = "This article explains options you have for defining the size of a Kubernetes cluster with Giant Swarm, and automatically scaling it"
+date = "2019-01-17"
+weight = 45
+type = "page"
+categories = ["basics"]
++++
+
+# Cluster Size and Autoscaling
+
+Starting with release version 6.3.0 (TODO) for AWS, you can leverage the benefits of the [Kubernetes autoscaling components](https://github.com/kubernetes/autoscaler) to define the number of worker nodes in a cluster based on demand.
```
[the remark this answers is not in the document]
marians: Will fix

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,60 @@
...
+When creating a cluster without specifying the number of worker nodes, **three** worker nodes will be created. This is also the minimal number supported by Giant Swarm. On AWS starting with release version 6.3.0, when not specified, the maximum number of worker nodes is also set to **three**.
+
+Technically, while you may be able to create and run smaller clusters successfully, we don't encourage this due to reduced resilience. We explicitly deactivate all sorts of alerts for clusters with less than three worker nodes and won't get notified in case of any problems.
+
+## Minimal worker node instance type
+
+Relying on autoscaling may have an effect on the instance type you should use for your cluster.
+To explain this more detail, it is helpful to explain how the autoscaler determines whether a node is needed or not.
+
+The autoscaler periodically calculates the utilization of a node based on CPU and memory requests, compared to the node's total capacity. The utilization is compared to a configurable _utilization threshold_, which is 50% by default. If the utilization is below the threshold, the autoscaler decides to remove the node. If it's above, it will keep the node.
```
calvix: Maybe I would somehow empathize that this is about kubernetes requests that are set on pods, otherwise, people might assume it's about the real utilization on the node which is not true. This then explains a little bit more than the issue with utilization over 50% as the critical components have set request to certain value but don't really have much real usage on an empty cluster.

src/content/basics/cluster-size-autoscaling/index.md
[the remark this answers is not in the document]
tuommaki: Third bullet is about scheduling hints and absolute resource usage limits that POD should be configured with. See this: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/

K8s scheduler uses `requests` to schedule PODs on available nodes and `limits` to provide absolute cgroup limits to running processes. These must be set correctly in order to have working workload scheduling in general, but when using `cluster-autoscaler`, it's even more important to have them correctly set.

src/content/basics/cluster-size-autoscaling/index.md
tuommaki: In other words: Utilization that `cluster-autoscaler` considers is the same as what k8s scheduler is working with. It's scheduling hints provided as `requests` settings in PODs and it's not directly connected to real CPU/memory utilization. When these scheduling hints don't represent the real workload that corresponding POD is producing, it results either in underutilization of HW resources (i.e. CPU is sitting idle or gigabytes of memory unused) or overloading of a node where the whole node goes unavailable due to load.

marians: Get it. Will add some info. This means we'll repeat ourselves compared to other articles, but that's fine.

calvix: also this PR https://github.com/giantswarm/kubernetes-cluster-autoscaler/pull/14 seems to change the value to 65 %

MarcelMue: Please don't fix this to 65% or something yet. I will sync with marian when we have a definitive final value.

src/content/basics/cluster-size-autoscaling/index.md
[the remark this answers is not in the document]
marians: I added an extra note regarding this last topic in my last commit.

marians:
Update for the reviewers: The cluster sizing page should make it very clear now that if autoscaling is desired, pods need resource requests.

I have also introduced some shortcodes, which allow us to specify an information in one place only, but use it in many places in text. This should simplify maintenance of values such as the utilization threshold.

src/content/basics/cluster-size-autoscaling/index.md
marians: This is now encoded in one place a so-called "shortcode" which is then re-used wherever necessary in docs.

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,66 @@
...
+Technically, while you may be able to create and run smaller clusters successfully, we don't encourage this due to reduced resilience. We explicitly deactivate all sorts of alerts for clusters with less than {{% minimal_supported_cluster_size_worker_nodes %}} worker nodes and won't get notified in case of any problems.
+
+## Minimal worker node instance type
+
+Relying on autoscaling may have an effect on the instance type you should use for your cluster.
+To illustrate this more detail, it is helpful to explain how the autoscaler determines whether a node is needed or not.
+
+The autoscaler periodically calculates the utilization of a node based on CPU and memory requests of all running pods, compared to the node's total capacity.
+
+**Note:** this means that your workloads need sensible CPU and memory requests in order to inform the autoscaler about the actual node utilization.
```
calvix: I would still rather use `pods` rather than `workloads` so specify  that we mean pod requests and not something else

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+The current utilization is compared to a configurable _utilization threshold_, which is {{% autoscaler_utilization_threshold %}} by default.
+If the utilization is below the threshold, the autoscaler decides to remove the node.
+If it's above, it will keep the node.
```
calvix: Here we describe what conditions need to be met in order to scale down, but we don't write anywhere about conditions that need to be met for scaling up.  It might even be worth separate paragraph so this info can be easily found without the need to read every word and I personally would never think to look for that info in `Minimal worker node instance type` section

src/content/basics/cluster-size-autoscaling/index.md
marians:
```suggestion
**Note:** this means that your pods need sensible CPU and memory requests in order to inform the autoscaler about the actual node utilization.
```

calvix: :+1:

src/content/basics/cluster-size-autoscaling/index.md
marians: @calvix I agree that the explanation on how the autoscaler works should be moved into its own section. Will cahnge.

@teemow There are several instances types in our supported list that might be too small for the default utilization threshold. c5.large, m3.large, m4.large, m5.large, r3.large, t3.large all have 2 cores only. And this list may vary. Also, our component's footprint may vary. So I wonder how we can keep this information useful AND maintainable. But for now I'd be fine with a list of instance types to stay away from.

marians: @teemow BTW have you seen that a bit further down the text contains `We recommend to use types with at least 4 CPU cores for autoscaling clusters.` ?

teemow: > have you seen that a bit further down the text

Yes that's why I said it would be nice if we would name instance types instead of how many CPUs. Nobody knows this without looking it up.

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,66 @@
...
+## Minimal and default cluster size
+
+When creating a cluster without specifying the number of worker nodes, {{% default_cluster_size_worker_nodes %}} worker nodes will be created. This is also the minimal number supported by Giant Swarm. On AWS starting with release version {{% first_aws_autoscaling_version %}}, when not specified, the maximum number of worker nodes is also set to {{% default_cluster_size_worker_nodes %}}.
+
+Technically, while you may be able to create and run smaller clusters successfully, we don't encourage this due to reduced resilience. We explicitly deactivate all sorts of alerts for clusters with less than {{% minimal_supported_cluster_size_worker_nodes %}} worker nodes and won't get notified in case of any problems.
+
+## Minimal worker node instance type
+
+Relying on autoscaling may have an effect on the instance type you should use for your cluster.
+To illustrate this more detail, it is helpful to explain how the autoscaler determines whether a node is needed or not.
```
mazzy89:
```suggestion
To illustrate this in more detail, it is helpful to explain how the autoscaler determines whether a node is needed or not.
```

src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+The autoscaler periodically calculates the utilization of a node based on CPU and memory requests of all running pods, compared to the node's total capacity.
+
+**Note:** this means that your pods need sensible CPU and memory requests in order to inform the autoscaler about the actual node utilization.
```
mazzy89: Does it mean that the user **must** set up requests (and perhaps limits) property in the YAML configuration? It is not clear here if any action is expected from the user.

tuommaki: Technically it's not mandatory, but scheduling doesn't work as expected and given the workloads we have experienced, in practice it is mandatory. Nevertheless, there's no validation nor similar mechanisms preventing from submitting ill-configured PODs. Hence this wording, I think.

marians:
Latest changes

- Added all the caveat info regarding the Ingress Controller replicas not being scaled.
- Moved description how the autoscaler works into its own section.

src/content/basics/cluster-size-autoscaling/index.md
marians: Thanks, fixed this in another commit.

src/content/basics/cluster-size-autoscaling/index.md
mazzy89: I see but we should probably enforce this by writing that. It's true that no validation for those kind of missing things (limits and requests) but maybe it worths to communicate this. we take two pigeons with one bean:

- customers will enforce requests and limits in their applications (which is generally a very good practice)
- cluster autoscaler will work smooth and properly

cc @marians

marians: I will try to make it clearer.

marians: Change note sentence to `This means that your pods MUST be configured with CPU and memory requests in order to inform the autoscaler about the actual node utilization.`. How about that?

mazzy89: Love that.

src/content/basics/cluster-size-autoscaling/index.md
marians: Considering this resolved. We tested with `c5.large`, which is the smallest type from our supported list, and scaling down worked. The information in the text should be sufficient.

PR #293, opened 2019-01-17
Fix case in headline

marians:


rossf7:
@marians Can this be merged? Looks like CI just failed due to the GitHub API rate limiting.

marians:
Thanks for the ping!

rossf7:
Np! I was getting to inbox 0 😄

PR #298, opened 2019-02-01
Align intermediate headline case

marians:
This changes all intermediate headlines to no longer use Title Case, as discussed internally.


marians:
The review request here is mainly FYI. I don't expect anyone to thoroughly review this.

PR #302, opened 2019-02-13
Add warning on configuring `use-proxy-protocol`

webwurst:
Goes with this https://github.com/giantswarm/kubernetes-nginx-ingress-controller/pull/79

### DIFF
--- a/src/content/guides/advanced-ingress-configuration/index.md
+++ b/src/content/guides/advanced-ingress-configuration/index.md
@@ -359,6 +359,8 @@ data:
   vts-default-filter-key: ""
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/guides/advanced-ingress-configuration/index.md
```diff
@@ -359,6 +359,8 @@ data:
   vts-default-filter-key: ""
 ```
 
+__Warning:__ We also allow setting `use-proxy-protocol: "true"/"false"`. This setting always applies globally for the `nginx-ingress-controller`. All applications providing services behind ingresses need to understand this protocoll or they will fail. Currently this setting probably only makes sense on bare metal installations, but will require a matching configuration on the load balancers.
```
puja108:
```suggestion
__Warning:__ We also allow setting `use-proxy-protocol: "true"/"false"`. This setting always applies globally for the `nginx-ingress-controller`. All applications providing services behind ingresses need to understand this protocol or they will fail. Furthermore, the load balancer in front of the ingress controller also needs to be set up correctly. So currently, customizing setting only makes sense on bare metal installations and will require a matching configuration on the load balancers.
```

PR #306, opened 2019-02-21
Audit policy

mazzy89:
Add small paragraph about Audit Policy

### DIFF
--- a/src/content/basics/kubernetes-on-giant-swarm/index.md
+++ b/src/content/basics/kubernetes-on-giant-swarm/index.md
@@ -42,6 +42,8 @@ As we are taking care of your cluster, we need to run some agents (e.g. for moni
 
 Similarly, some parts of the DNS, Ingress Controller, and Calico setups are visible to you inside your cluster. To ensure that your cluster runs without problems, please refrain from manipulating the `kube-system` namespace as well as the pods and other resources running in them if they are not documented.
 
+We customized the audit policy file by introducing rules that make the `Kube-apiserver` less verbose. The main philosophy behind is to mute those rules which have low-risk and produce high-volume of logs.  Check the repository [giantswarm/k8scloudconfig](https://github.com/giantswarm/k8scloudconfig) for a file named `audit-policy.yaml` under the folder named `policies`.
+  
 ### Specifics on AWS
 
 On AWS all resources (besides minor resources on S3 and KMS) pertaining to a cluster carry the two following tags:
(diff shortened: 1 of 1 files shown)


calvix:
maybe at least  show an example of a path with variables as version as `v_X_Y/files/....`

mazzy89:
I know, unfortunately, we already have discussed this here https://gigantic.slack.com/archives/C7865GLSY/p1550655658010100 but this seems the most appropriate solution for the moment. Any proposals?

mazzy89:
> maybe at least show an example of a path with variables as version as `v_X_Y/files/....`

Yes that it does not look a bad idea. I like that.

src/content/basics/kubernetes-on-giant-swarm/index.md
```diff
@@ -42,6 +42,8 @@ As we are taking care of your cluster, we need to run some agents (e.g. for moni
 
 Similarly, some parts of the DNS, Ingress Controller, and Calico setups are visible to you inside your cluster. To ensure that your cluster runs without problems, please refrain from manipulating the `kube-system` namespace as well as the pods and other resources running in them if they are not documented.
 
+We customized the audit policy file by introducing rules that make the `Kube-apiserver` less verbose. The main philosophy behind is to mute those rules which have low-risk and produce high-volume of logs.  Check the manifest `audit-policy.yaml` in the repository [giantswarm/k8scloudconfig](https://github.com/giantswarm/k8scloudconfig) under `v_X_Y_Z/files/policies`.
```
marians:
```suggestion
We customize the audit policy file to eliminate rules which are both low-risk and produce a high volume of log entries. For full details, check the manifest `audit-policy.yaml` in the repository [giantswarm/k8scloudconfig](https://github.com/giantswarm/k8scloudconfig) under `v_X_Y_Z/files/policies`.
```

Suggesting some rephrasing

marians:
I think this will work. Cool!

As a general note: Adding a single line break after each sentence in markup would simplify reviews and especially making suggestions to changes. It will also make future diffs more readable.

This is something we don't have consistently in our content, so it's absolutely not your fault.

src/content/basics/kubernetes-on-giant-swarm/index.md
mazzy89: So we won't go with the convention `v_X_Y_Z` as proposed above by Vaclav?

marians: Yes, we should. I left that part untouched.

mazzy89: Oh right sorry. I was confused by GH

PR #311, opened 2019-03-11
Add documentation for default cert and annotation prefix customization

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/5490

### DIFF
--- a/src/content/guides/advanced-ingress-configuration/index.md
+++ b/src/content/guides/advanced-ingress-configuration/index.md
@@ -348,6 +348,7 @@ data:
   enable-vts-status: "true"
   error-log-level: "error"
   hsts: "false"
+  large-client-header-buffers: "4 8K"
   log-format-upstream: "$status $body_bytes_sent $http_referer"
   server-name-hash-bucket-size: "1024"
   server-name-hash-max-size: "1024"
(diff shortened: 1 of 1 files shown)


src/content/guides/advanced-ingress-configuration/index.md
```diff
@@ -348,6 +348,7 @@ data:
   enable-vts-status: "true"
   error-log-level: "error"
   hsts: "false"
+  large-client-header-buffers: "4 8K"
```
rossf7: Is the `4 8K` a typo? Otherwise I'm missing something :)

src/content/guides/advanced-ingress-configuration/index.md
```diff
@@ -363,6 +364,24 @@ __Warning:__ We also allow setting `use-proxy-protocol: "true"/"false"`. This se
...
+Once you want to have the default server on the nginx controller support TLS you need to provide a certificate. This is made thanks to the flag `--default-ssl-certificate`. Now you can provide this value in the `user-value` configmap to force the component to be restarted with the certificate provided. The value of the property should be the namespace and secret name which holds the certificate content.
+
+```yaml
+data:
+  default-ssl-certificate: "default/foo-tls"
+```
+
+## Custom annotation prefix
+
+By default we use the default annotation prefix `nginx.ingress.kubernetes.io` in the ingress controller. In case the customer needs to have a especific one it can be done through the 'user-values' configmap. It is recommended when there is more than one ingress controller, so in the ingress resource the prefix can discriminate between controllers.
```
rossf7:
```suggestion
By default we use the standard annotation prefix `nginx.ingress.kubernetes.io` in the ingress controller. In case the customer needs to have a specific one this can be done via the 'user-values' configmap. This is recommended when there is more than one ingress controller. So in the ingress resource the prefix can be used to distinguish between controllers.
```

WDYT to using standard annotation prefix? The use of default twice made this more confusing for me.

src/content/guides/advanced-ingress-configuration/index.md
```diff
@@ -363,6 +364,24 @@ __Warning:__ We also allow setting `use-proxy-protocol: "true"/"false"`. This se
 
 On cluster creation the ConfigMap is empty and above defaults will be applied to the final Ingress Controller deployment. To override any of the above values, you just need to add the respective line in the data field of the user ConfigMap.
 
+## Default certificate
+
+Once you want to have the default server on the nginx controller support TLS you need to provide a certificate. This is made thanks to the flag `--default-ssl-certificate`. Now you can provide this value in the `user-value` configmap to force the component to be restarted with the certificate provided. The value of the property should be the namespace and secret name which holds the certificate content.
```
rossf7:
```suggestion
When you want to have the default server on the nginx controller support TLS you need to provide a certificate. This is configured using the flag `--default-ssl-certificate`. Now you can provide this value in the `user-value` configmap to force the component to be restarted with the provided certificate. The value of the property should be the namespace and secret name which holds the certificate content.
```

src/content/guides/advanced-ingress-configuration/index.md
pipo02mix: No. First number is the number and second the size.

http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers

rossf7: Ah cool. Got it now, thx!

PR #319, opened 2019-05-09
Upgrade nginx base image

marians:
to fix a vulnerability

https://quay.io/repository/giantswarm/docs/manifest/sha256:e0461a5b5094d4df57e365b3351198559b084a65798641ca47cfc5d2e27794d3?tab=vulnerabilities&fixable=true

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.14.1-alpine
+FROM nginx:1.15.12-alpine
 
 EXPOSE  80
 WORKDIR /


marians:
This doesn't actually fix the problem, but it keeps us more up to date.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nginx:1.14.1-alpine
+FROM nginx:1.15.12-alpine
```
tuommaki: Out of curiosity: How does this `nginx` container differ from having a "standard" alpine container with nginx installation step?

marians: No idea. We use `nginx` for convenience reasons. Feel free to inspect the image's Dockerfile for details. https://github.com/nginxinc/docker-nginx/blob/e5123eea0d29c8d13df17d782f15679458ff899e/mainline/alpine/Dockerfile

PR #326, opened 2019-06-12
Docs front page: switch user guide lists

marians:
Towards: https://github.com/giantswarm/giantswarm/issues/6001

### Before:

Screenshot 2019-06-12 10 12 50(2)

### After

image

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,7 +2,7 @@ version: "2"
 services:
 
   content:
-    build: .
+    # to build this image locally, execute 'make'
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: "2"
 services:
 
   content:
-    build: .
+    # to build this image locally, execute 'make'
     image: quay.io/giantswarm/docs:latest
```
marians: This change is unrelated. I figured that it is misleading to have the build key here, as the `make` command is required.

PR #332, opened 2019-07-12
Adjust screenshots

oponder:
This adjusts the screenshot on the reference page, and adds a bit of a teaser screenshot to the overview page.

The reason I thought to go with a teaser is because screenshots will become outdated soon, and I didn't think it would be productive to go into a discussion about which apps to show and which not.


oponder:
Intro teaser image looks like: 
Screenshot 2019-07-12 at 8 40 48 PM

PR #334, opened 2019-07-16
First Version of Azure Achitecture

othylmann:
The pictures are still to be done. I will have to search for those. I already renamed the files (replace azure with aws to find the original ones) but will likely have to talk to Marat or something to build new ones.


src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+Giant Swarm's Architecture is split into two logical parts, one being the Control Plane and the other being multiple Tenant Clusters. Among other services (for details see Service Architecture below), the control plane runs our [Azure Operator](https://github.com/giantswarm/azure-operator), which handles the full lifecycle management of Tenant Clusters.
+
+## Giant Swarm control plane
+
+![Control Plane Architecture](/img/architecture-azure-control-plane.png)
+
+The above diagram shows our control plane setup running with two bastion hosts in different availability zones (???). All instances are in a private subnet and only accessible through the bastion hosts. Bastions hosts are available via Azure VPN connection only.
+
+The cluster has several APIs and Interfaces. The Kubernetes API of the Control Plane is only available to Giant Swarm operations personel and only through the Bastion hosts.
```
rossf7:
```suggestion
The cluster has several APIs and Interfaces. The Kubernetes API of the Control Plane is only available to Giant Swarm operations personnel and only through the Bastion hosts. 
```

Just a typo that is in the AWS version too.

src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+# The Giant Swarm Azure Architecture
+
+Giant Swarm's Architecture is split into two logical parts, one being the Control Plane and the other being multiple Tenant Clusters. Among other services (for details see Service Architecture below), the control plane runs our [Azure Operator](https://github.com/giantswarm/azure-operator), which handles the full lifecycle management of Tenant Clusters.
+
+## Giant Swarm control plane
+
+![Control Plane Architecture](/img/architecture-azure-control-plane.png)
+
+The above diagram shows our control plane setup running with two bastion hosts in different availability zones (???). All instances are in a private subnet and only accessible through the bastion hosts. Bastions hosts are available via Azure VPN connection only.
```
rossf7: Yes I think Azure uses different terminology instead of AZs. I can check.

src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+The cluster has several APIs and Interfaces. The Kubernetes API of the Control Plane is only available to Giant Swarm operations personel and only through the Bastion hosts. 
+
+The Giant Swarm API, Monitoring and Alerting frontends as well as our Web Management Interface, Happa, are exposed through the Control Plane Ingress Controller, which sits behind a public load balancer (???).
```
rossf7: Same here I can check on the LB.

src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Giant Swarm tenant cluster
+
+![Tenant Cluster Architecture](/img/architecture-azure-tenant-cluster.png)
+
+Via the Giant Swarm API, our [CLI](https://github.com/giantswarm/gsctl), or our Happa interface, you can start Tenant Clusters of different sizes. There's a selection of recommended Azure instance types (???), which can be adjusted if needed. 
+
+Each cluster resides in its own VPC (WHAT IS IT IN AZURE?). All Azure machines are in their own private subnet ([Networking Deepdive](https://blog.giantswarm.io/deep-dive-into-kubernetes-networking-in-azure/)). There are two possible access routes into the cluster.
```
rossf7: I think its Virtual Network but I'll check. Why can't they use the same naming as AWS!

Great to link to Roma's post. <3

othylmann: Virtual Network fine, just not a VPC. VPC as a term does not exist. I think. I am glad to be proven wrong :)

src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+# The Giant Swarm Azure Architecture
+
+Giant Swarm's Architecture is split into two logical parts, one being the Control Plane and the other being multiple Tenant Clusters. Among other services (for details see Service Architecture below), the control plane runs our [Azure Operator](https://github.com/giantswarm/azure-operator), which handles the full lifecycle management of Tenant Clusters.
+
+## Giant Swarm control plane
+
+![Control Plane Architecture](/img/architecture-azure-control-plane.png)
+
+The above diagram shows our control plane setup running with two bastion hosts in different availability zones. All instances are in a private subnet and only accessible through the bastion hosts. Bastions hosts are available via Azure VPN connection only.
```
rossf7: > The above diagram shows our control plane setup running with two bastion hosts in different availability zones.

I'm not sure about the AZ part here. @kopiczko can you help with this?

kopiczko: AFAICT we don't use AZs in CP. We only use availability sets (AS) which according to what I linked here is something like separate rack. We use 3 fault domains. Masters, workers and bastions are in separate ASs. https://cloudopszone.com/availability-sets-and-availability-zones/

For TCs I think we don't even use availability sets.

rossf7: Thanks for this. The control plane diagram got updated to show the fault domains. The tenant cluster doesn't have this but neither does the AWS diagram. So I think that's fine for the first version.

src/content/basics/azure-architecture/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+Via the Giant Swarm API, our [CLI](https://github.com/giantswarm/gsctl), or our Happa interface, you can start Tenant Clusters of different sizes. There's a selection of recommended Azure instance types, which can be adjusted if needed. 
+
+Each cluster resides in its own Virtual Network. All Azure machines are in their own private subnet ([Networking Deepdive](https://blog.giantswarm.io/deep-dive-into-kubernetes-networking-in-azure/)). There are two possible access routes into the cluster. 
+
+One is the Kubernetes API that can be connected to your Identity Management System using OIDC. The other is the Ingress Controller (exposed through a public Load Balancer), with which you can expose services running inside your cluster publicly. The services will be mapped to domains handled through Azure DNS. For the final URL, you can point a CNAME to the Ingress URL. 
+
+## Service architecture
+
+![Service Architecture](/img/architecture-aws-services.png)
```
kopiczko: This is so weird that there is `aws` in the path.

src/content/basics/azure-architecture/index.md
kopiczko: The image may suggest that the bastions/masters/workers are collocated in the same fault domains which is not true. I don't know if this is intentional compromise to make the image clearer. Otherwise LGTM.

src/content/basics/azure-architecture/index.md
rossf7: Agree but this is how it's done in the onprem docs as well. I think we should split but I think that can be done in a later PR.

src/content/basics/azure-architecture/index.md
rossf7: Yes it makes the image clearer. Going to go with this for the first version.

PR #335, opened 2019-07-17
Improve arch diograms

rossf7:
Follow up to https://github.com/giantswarm/docs/pull/334.

Some cleanup that @morotki helped me with. <3

Now we have separate services diagrams for each provider. Also replaces all uses of Guest Cluster with Tenant Cluster. 🎉

The war against guest clusters continues. 😄


src/layouts/index.html
```diff
@@ -31,6 +31,9 @@ <h2>Learn about Giant Swarm</h2>
         </li>
       </ul>
     </div>
+  </div>
```
rossf7: Fixing the markup to tidy up the docs home page.

JosephSalisbury:
And while it's in my brain - we can remove the domain validation / couchdb parts, those are so old

rossf7:
> we can remove the domain validation / couchdb parts, those are so old

@JosephSalisbury Cool thanks I'll do a follow up for that.

PR #348, opened 2019-09-17
Add digital climate strike script

marians:
This PR adds a script that

- before September 20 will add a banner to our site
- on September 20 will add an overlay

as described at https://digital.globalclimatestrike.net/

See also https://github.com/giantswarm/giantswarmio-webapp/pull/593


ograu:
💚

PR #352, opened 2019-09-25
Add general node pools page

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5465 https://github.com/giantswarm/giantswarm/issues/7075

This PR adds a page at https://docs.giantswarm.io/basics/nodepools/ that explains all the basics about Node Pools.

PR #362, opened 2019-10-28
Add gsctl reference pages for node pools

marians:
For https://github.com/giantswarm/giantswarm/issues/7075

This PR adds the missing reference pages for new gsctl commands introduced for node pools.

It links to the /basics/nodepools/ page which is introduced in PR #352 and therefore the linkcheck still fails.


marians:
Thanks!

PR #373, opened 2019-11-06
Move common local assets to server

axbarsan:
Linked to giantswarm/web-assets#40


marians:
What was the problem here @axbarsan ?

PR #377, opened 2019-11-11
Move common local assets to server

axbarsan:
Addresses giantswarm/giantswarm#7530


marians:
Link checks were failing because web-assets was not up-to-date yet. I redeployed web-assets, re-ran the last build, now link checks are fine.

PR #383, opened 2019-11-21
Update to Azure architecture on docs

Oshratn:
Removed references to Happa (as a full sweep of docs, since the term needs to go away from customer-facing documentation).
Additions for purposes of clarity and readability.

Towards #7587

### DIFF
--- a/src/content/basics/azure-architecture/index.md
+++ b/src/content/basics/azure-architecture/index.md
@@ -5,7 +5,7 @@ date = "2019-11-21"
 weight = 20
 type = "page"
 categories = ["basics"]
-last review date = "2019-11-25"
+last_review_date = "2019-11-26"
 +++
 
 # The Giant Swarm Azure Architecture
--- a/src/content/basics/azure-architecture/index.md
+++ b/src/content/basics/azure-architecture/index.md
@@ -5,6 +5,7 @@ date = "2019-11-21"
 weight = 20
 type = "page"
 categories = ["basics"]
+last review date = "2019-11-25"
 +++
 
 # The Giant Swarm Azure Architecture
--- a/src/content/basics/azure-architecture/index.md
+++ b/src/content/basics/azure-architecture/index.md
@@ -1,7 +1,7 @@
 +++
 title = "The Giant Swarm Azure Architecture"
 description = "Architecture Overview showing how Giant Swarm is set up on Microsoft Azure"
-date = "2019-07-16"
+date = "2019-11-21"
 weight = 20
 type = "page"
 categories = ["basics"]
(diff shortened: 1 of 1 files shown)


src/content/basics/azure-architecture/index.md
```diff
@@ -35,10 +35,12 @@ One is the Kubernetes API that can be connected to your Identity Management Syst
 
 ![Service Architecture](/img/architecture-azure-services.png)
 
-To make your life easier, we have developed a lot of different services within our Control Plane that allow both our operations team and you as users of our API and interfaces to easily manage Kubernetes clusters. Most of these services should be self explanatory. 
+To make your life easier, we have developed a lot of different services within our Control Plane that allow both our operations team and you as, users of our API and interfaces to easily manage Kubernetes clusters. Most of these services should be self explanatory, due to the level of granularity we maintain.
```
rossf7: Sorry but the addition of `due to the level of granularity we maintain.` makes this less clear for me.

I think the meaning about being self explanatory is because the terms used are common in this space.

Oshratn: The thing is that the existing text says 'should be self-explanatory' and I want to add the reasoning as to why it's self-explanatory, because what happens if an individual does not feel it is indeed self-explanatory. 
The addition was my attempt to allude to the fact that we keep a separation of concerns, so each operator does one thing, which is why it is self-explanatory (Whew, that was long. Maybe we need to iron this out verbally).

rossf7: OK that makes sense but if the audience is other architects they should understand those terms.

But I'm also fine with just removing self explanatory. @piontec you have fresher eyes than me. WDYT?

src/content/basics/azure-architecture/index.md
```diff
@@ -1,10 +1,11 @@
 +++
 title = "The Giant Swarm Azure Architecture"
 description = "Architecture Overview showing how Giant Swarm is set up on Microsoft Azure"
-date = "2019-07-16"
+date = "2019-11-21"
 weight = 20
 type = "page"
 categories = ["basics"]
+last review date = "2019-11-25"
```
rossf7:
```suggestion
last_review_date = "2019-11-25"
```

I think this is a great addition but it actually breaks the automated tests! Can you try it with this change?

Oshratn: My plan is to add that - slowly - to all the documents, in order to add sanity to the review cycle. If it breaks something then we'll have to think of something else.

rossf7: @Oshratn The problem is the spaces in `last review date` please try it again with underscores like in my suggestion. That should fix it.

Oshratn: Gotcha, will do.

Oshratn:
Done

Oshratn:
@piontec will you get to this or should I merge as-is?

rossf7:
The automated tested are fixed 🎉 but now there is a merge conflict. 👎 

@Oshratn Let me know if you want a hand with merging this.

PR #387, opened 2019-12-02
Node pools: add info on autoscaling behaviour

marians:
For https://github.com/giantswarm/giantswarm/issues/7075

This adds a little section on autoscaling to the node pools page.

### DIFF
--- a/src/content/basics/nodepools/index.md
+++ b/src/content/basics/nodepools/index.md
@@ -145,7 +145,7 @@ With node pools, you set the autoscaling range per node pool. The Kubernetes clu
 
 If you assign workloads to node pools as described [above](#assigning-workloads) and the autoscaler finds pods in `Pending` state, it will decide based on the node selectors which node pools to scale up.
 
-In case there are workloads not assiged to any node pools, the autoscaler may pick any node pool for scaling. For details on the decision logic, pelaase check the upstream [FAQ for AWS](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md).
+In case there are workloads not assigned to any node pools, the autoscaler may pick any node pool for scaling. For details on the decision logic, please check the upstream [FAQ for AWS](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md).
 
 ## Limitations
 
--- a/src/content/basics/nodepools/index.md
+++ b/src/content/basics/nodepools/index.md
@@ -1,7 +1,7 @@
 ---
 title: Node Pools
 description: A general description of node pools as a concept, it's benefits, and some details you should be aware of.
-date: 2019-11-19
+date: 2019-12-02
 weight: 130
 type: page
 categories: ["basics"]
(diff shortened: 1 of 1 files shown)


src/content/basics/nodepools/index.md
```diff
@@ -139,6 +139,14 @@ Just as the Giant Swarm API schema for v4 (without node pools) and v5 (with node
 The new definition schema for v5 allows for defining cluster and node pool details in one file,
 to be submitted for creation via the [`gsctl create cluster`](/reference/gsctl/create-cluster/) command.
 
+## Node pools and autoscaling {#autoscaling}
+
+With node pools, you set the autoscaling range per node pool. The Kubernetes cluster autoscaler has to decide which node pool to scale under which circumstances.
```
pipo02mix:
```suggestion
With node pools, the autoscaling is configured per node pool automatically. The Kubernetes cluster autoscaler has to decide which node pool to scale under which circumstances.
```
In reality customer does not perform any action

src/content/basics/nodepools/index.md
```diff
@@ -139,6 +139,14 @@ Just as the Giant Swarm API schema for v4 (without node pools) and v5 (with node
...
+
+If you assign workloads to node pools as described [above](#assigning-workloads) and the autoscaler finds pods that are not assigned to any node, it will decide based on the node selectors which node pools to scale up.
```
pipo02mix:
```suggestion
If you assign workloads to node pools as described [above](#assigning-workloads) and the autoscaler finds pods in `Pending` state, it will decide based on the node selectors which node pools to scale up.
```

src/content/basics/nodepools/index.md
```diff
@@ -139,6 +139,14 @@ Just as the Giant Swarm API schema for v4 (without node pools) and v5 (with node
...
+
+In case there are workloads not assiged to any node pools, the autoscaler may pick any node pool for scaling. For details on the decision logic, pelaase check the upstream [FAQ for AWS](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md).
```
pipo02mix:
```suggestion
In case there are workloads not assigned to any node pools, the autoscaler may pick any node pool for scaling. For details on the decision logic, please check the upstream [FAQ for AWS](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md).
```

src/content/basics/nodepools/index.md
marians: Actually, yes. The customer initially decides in which range the node pool can scale.

pipo02mix: in which range? can you explain it a bit more? what do you mean by range?

marians: I mean the number of nodes, which is configured by the user as a MIN and MAX number.

marians:
Thanks for the input! Is this acceptable like it is?

src/content/basics/nodepools/index.md
pipo02mix: AH ok I was thinking from the ASG configuration side

PR #390, opened 2019-12-19
Node pools release update

marians:
- Remove disclaimers stating that node pools is coming soon
- Add gsctl commands that are purely dealing with node pools to the reference index
- Add some links to the node pools basics page


marians:
FYI @xh3b4sd

PR #391, opened 2019-12-27
Update index.md

snizhana-dynnyk:
fixes broken links to images


marians:
Thanks for the fix!

FYI: I'd prefer us to change image locations consistently to reside in the same folder where the markdown resides they are referenced in.

This hasn't been possible with HUGO when we began, but it is now. It would simplify things for us.

snizhana-dynnyk:
Oh, thanks @marians! Now it makes sense why tests didn't pass, I closed PR because there was no one around to ask but now I can fix this correctly :)

PR #400, opened 2020-01-28
Changes AWS architecture page on docs.giantswarm

Oshratn:
Towards Issue https://github.com/giantswarm/giantswarm/issues/8626

Added closing sentence. 
Made some changes to English

Please verify that the rest of the document is correct.


J-K-C:
@cornelius-keller would you be able to check this on behalf of me for @Oshratn please? I think you're probably better suited to check it. I've read it too and from my perspective it sounds right however, probably best to get your input as well to be sure. thank you.

cornelius-keller:
Hi @Oshratn, @J-K-C as far as I see it the illustration of the tenant cluster architecture is not reflecting node pools yet. I don't know if it is in the scope of this PR but I think we should update it in the future.

Oshratn:
If moving forward this is the default, I would change it now. Or maybe we can use current illustration to say it is deprecated and will be moving out of support and provide the newer illustration.

Please provide the correct illustration and I will figure out how to work it in.

cornelius-keller:
I created a diagram for the aws tenant cluster architecture here
https://github.com/giantswarm/giantswarm/blob/master/architecture/aws/aws.md
a while ago. I don't know if the quality is good enough for the website,
but it can serve as a source for new more professionally designed
illustrations if we need them.

Oshratn:
I'll double check that we have the correct diagram. Snizhana updated the
diagram not many weeks ago.

Oshratn:
Diagrams will be reviewed as a whole via issue https://github.com/giantswarm/giantswarm/issues/9549

PR #407, opened 2020-02-05
Reduce number of linkchecker threads.

oponder:
Currently the CI is flapping due to failing linkchecker. 

Testing if reducing the number of threads will produce a more consistent result.


marians:
Another thing we can try: increase timeouts. For example: 

    --timeout=100

See https://linkchecker.github.io/linkchecker/man1/linkchecker.1.html

PR #412, opened 2020-02-17
Update docs on minimal instance type in Cluster Size and Autoscaling article

marians:
Part of https://github.com/giantswarm/giantswarm/issues/8748

This PR adapts the docs to eliminate the smaller `*.large` instance types everywhere, which cannot be used any more.

The basic Autoscaling article has been stripped a bit to remove some historical info and to simplify the content.

### DIFF
--- a/src/content/guides/prepare-aws-account-for-tenant-clusters/index.md
+++ b/src/content/guides/prepare-aws-account-for-tenant-clusters/index.md
@@ -64,7 +64,7 @@ These are the limit increases to be requested, grouped by limit type:
   - Auto Scaling Groups per region: **250**
   - Launch Configurations per region: **500**
 - EC2 Instances
-  - m3.large per region: **250**
+  - m4.xlarge per region: **250**
   - m4.2xlarge per region: **250**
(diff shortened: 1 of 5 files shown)


src/content/guides/prepare-aws-account-for-tenant-clusters/index.md
```diff
@@ -64,7 +64,7 @@ These are the limit increases to be requested, grouped by limit type:
   - Auto Scaling Groups per region: **250**
   - Launch Configurations per region: **500**
 - EC2 Instances
-  - m3.large per region: **250**
```
marians: @calvix I think in the past at some point `m3.large` was a default instance type. Now the default is `m4.xlarge`.

Can you make sure there is no type missing which we use frequently?

calvix: tbh this heavily depends on customers instance type

marians: OK. I added an extra sentence about that.

PR #417, opened 2020-02-26
Move linkcheck into a separate job

marians:
This PR attempts to make the linkcheck step optional and not be executed when releasing an app.

As a result, the other checks are much faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
 workflows:
   version: 2
 
-  build_and_linkcheck:
+  build-and-linkcheck:
     jobs:
       - build
       - linkcheck:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,21 @@ orbs:
 
 
 jobs:
+  linkcheck:
+    machine:
+      image: "ubuntu-1604:201903-01"
+
+    steps:
+      - run:
+          name: Link check
+          command: |
+            docker run -d --rm --name docs -p 8080:8080 quay.io/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:$CIRCLE_SHA1
+            docker run --rm -ti --name linkchecker \
+              --link docs:docs \
+              linkchecker/linkchecker \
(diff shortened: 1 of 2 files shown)


marians:
Reviewers: don't get distracted by the fact that one check failed. That's the link check, currently failing because one web site times out.

.circleci/config.yml
```diff
@@ -58,6 +63,15 @@ jobs:
             fi
 
 workflows:
+  version: 2
+
+  build_and_linkcheck:
```
tfussell: I think we usually use hyphens to separate multi-word job/command/workflow names in CircleCI.

marians: PTAL

.circleci/config.yml
```diff
@@ -58,6 +63,15 @@ jobs:
             fi
 
 workflows:
+  version: 2
+
+  build-and-linkcheck:
```
tfussell: Instead of two workflows, I think I'd prefer to see this as a single workflow. This workflow is usually called `build` in our repos. You'd just need to combine the `package-and-push-chart-on-tag` and `build-and-linkcheck` jobs into one. Here's what this looks like in the CircleCI UI https://circleci.com/workflow-run/8e10efe4-4379-48a4-9cb4-d5f2296d8490 This keeps the build history cleaner and lets us add more steps as needed.

.circleci/config.yml
tfussell: Nice!

.circleci/config.yml
marians: Is that compatible with one of my goals, being that `package-and-push-chart-on-tag` does not trigger the link check again?

tfussell: Yes, as far as I know. You build up a tree structure of jobs that depend on other jobs using `requires`. The package-and-push job and the linkcheck job are independent so you make both of them depend on the build job and your goals should be satisfied. `Require status checks to pass before merging` isn't enabled for this repo so it shouldn't affect merging either.

marians:
I'll have to save @tfussell's suggestion for some time in the future, to be able to move on here

PR #418, opened 2020-02-28
Remove confusing line from AWS account prep doc

glitchcrab:
Looks like I missed this when updating this page.


glitchcrab:
I'm gonna merge this without a review as it's only a tiny change and should have been part of my previous PR.

marians:
Thanks!

PR #420, opened 2020-03-05
Add Arch Linux gsctl install instructions

glitchcrab:


glitchcrab:
Just a note that there are more complicated ways of installing AUR packages on Arch which avoid the use of an AUR helper, but I didn't include them as I think anyone who preferred to do that wouldn't need any guidance (beyond knowing that the package is available).

PR #422, opened 2020-03-11
Mention gsctl support for using cluster names as identifiers

axbarsan:
Related to https://github.com/giantswarm/gsctl/issues/188

We should mention being able to use cluster names with gsctl commands, now that it's available for all of them 👍

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ sitesearch-volume-data
 kubernetes/secrets.yaml
 /.vscode
 /volumes
+/.idea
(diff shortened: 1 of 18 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ sitesearch-volume-data
 kubernetes/secrets.yaml
 /.vscode
 /volumes
+/.idea
```
oponder: Perhaps add this to a global gitignore? https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

I don't mind at all though.

axbarsan: I didn't know about that. I've added it, thank you!

I'll still keep `.idea` here, though, since `.vscode` is also there. Code editor equality!

PR #428, opened 2020-03-18
Add Azure to Multi-az documentation page

MartaD:
Please review "Details you should know about" section and suggest missing/incorrect statements.


MartaD:
> I didn’t know about this when I updated the docs 😅

Yeah, there was a task (still in progress on our board but somehow not visible!) with link to this page , I got pinged by stalebot 😊 thanks for quick review!

PR #433, opened 2020-03-31
Edited for readability

Oshratn:
Updated dates and did some light editing for readability.


Oshratn:
Towards Issue #9644.
Still open question is whether we can remove reference to GS 6.1.0?

marians:
Where is that reference? I can't see it.

Oshratn:
Sheesh, I saw it today, but since it's not there, then the question is irrelevant.
I guess we can just proceed with the review of the editing.

PR #435, opened 2020-03-31
Review and Edit

Oshratn:
Towards Issue #9647.

Reviewed and edited for readability. 

I have concerns with regards to how current the document itself is due to recent work done on HPA.

### DIFF
--- a/src/content/basics/cluster-size-autoscaling/index.md
+++ b/src/content/basics/cluster-size-autoscaling/index.md
@@ -5,6 +5,7 @@ date = "2020-02-17"
 weight = 120
 type = "page"
 categories = ["basics"]
+last-review-date = "2020-03-31"
 +++
 
 # Cluster Size and Autoscaling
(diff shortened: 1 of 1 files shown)


src/content/basics/cluster-size-autoscaling/index.md
```diff
@@ -5,6 +5,7 @@ date = "2020-02-17"
 weight = 120
 type = "page"
 categories = ["basics"]
+last-review-date = "2020-03-31"
```
oponder, line 8: oh, just noticed this.. is this a new thing we are doing? I think this field doesn't really get used programatically in any way, is it just to keep a note for yourself when you last saw it?

Any reason we dont can't `date` above for that?

Oshratn: It is indeed a new thing and I am adding it for internal purposes. Sometimes a doc doesn't need to be edited, but it does require a review every once in a while.

Oshratn:
> Changes look good to me, not sure if they reflect the reality of HPA now, but as a general review and edit, looks good to me.

If it doesn't reflect HPA it should, since I have a feeling it is live.

Oshratn:
@cokiengchiara @sslavic this is a double ping. I'd like to ensure that this is a good snapshot for where we are now.
Please let's have a short meeting if HPA is live and requires a dramatic change in this document.

PR #439, opened 2020-04-06
Document that it is possible to change app versions using the web int…

oponder:
Towards finishing PDV in: https://github.com/giantswarm/giantswarm/issues/8022


oponder:
Going with a ping @giantswarm/team-batman

PR #440, opened 2020-04-07
Name Change: App Catalog to App Platform

Oshratn:
In the ramp towards that release of app catalog, we have revisited naming conventions and in order to avoid confusion, an App Catalog - is a group of curated apps, whereas the App Platform is the enabler of the App Catalogs.


oponder:
Name change sounds ok to me!

But there'll be quite a few more places in these docs,  website, and possible Happa, where we have to make the change.

Oshratn:
This conflicts with #441 @teemow / @puja108 help please.

puja108:
As it's a single word change the easiest would be to redo your change and close this

PR #442, opened 2020-04-07
Define the version of apiextensions we want to build crd docs from.

oponder:
Using the new flag, this way the build should become more deterministic. (we should also probably not use `latest` of crd-docs-generator, but im ok with that for now)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ build: vendor build-css
 	# Generate the Control Plane K8s API reference documentation.
 	docker run \
 		-v ${PWD}/build/content/reference/cp-k8s-api:/opt/crd-docs-generator/output \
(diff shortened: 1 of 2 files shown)


marians:
I am somehow reluctant here. It means that whenever we want to pick up new CRDs, we cannot simply tag a docs release. Instead we have to

1. (if not done yet) tag apiextensions
2. Modify the tag in docs
3. Tag a new docs release

oponder:
True... it's more steps, but it does mean we will not accidentally tag a version of docs that can't be built.

marians:
Why not keep/make crd-docs-generator tolerant to errors?

oponder:
And it will prevent people like Oshrat (and myself) who are making other edits to docs, from getting a failed build just because of something incorrect in `apiextensions` master.

PR #444, opened 2020-04-07
App Config Reference improvements

marians:
Describing all these changes would take some time. I would be happy to explain them in a call.

The goal was to make the page easier to read, understand, more consistent with other pages, make URLs more stable and much more.

### DIFF
--- a/src/content/reference/app-configuration/index.md
+++ b/src/content/reference/app-configuration/index.md
@@ -1,7 +1,7 @@
 ---
 title: App Configuration
 description: Documentation on the various levels of App configuration and how they get merged into a final Values object.
-date: 2020-04-07
+date: 2020-04-08
 type: subsection
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/reference/app-configuration/index.md
```diff
@@ -160,52 +155,48 @@ data:
...
+|||`.spec.userConfig.secret.namespace`|
 
-When setting user level configuration using the Giant Swarm API or our Web Interface,
-the fields in the `App CR` are edited automatically for you while creating
+When setting user level configuration using the Giant Swarm API or our [web interface](/reference/web-interface/),
+the fields in the App CR are edited automatically for you while creating
 the `ConfigMap` or `Secret`.
 
-## What is the right format for values in the ConfigMap and Secret?
+## Encoding values in ConfigMap and Secret {#encoding}
```
oponder, line 190: Not sure about the heading having the word 'encoding' in it.

The values for secrets have to be encoded, but that is almost kubernetes storage detail.

The key under which we expect your configuration (`.data.values`) is not an encoding.

marians: OK, changed

PR #456, opened 2020-04-16
Change default keypair/kubeconfig TTL to 1 day

axbarsan:
To be in line with https://github.com/giantswarm/giantswarm/issues/4630

### DIFF
--- a/src/content/reference/gsctl/create-keypair.md
+++ b/src/content/reference/gsctl/create-keypair.md
@@ -29,7 +29,7 @@ $ gsctl create keypair --cluster "Cluster name"
 The output will look similar to the one below:
 
 ```nohighlight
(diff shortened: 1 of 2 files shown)


src/content/reference/gsctl/create-keypair.md
```diff
@@ -29,13 +29,13 @@ $ gsctl create keypair --cluster "Cluster name"
 The output will look similar to the one below:
 
 ```nohighlight
-New key pair created with ID 114e2de892b2dcfdf96c92f3c69f39e564ecec1e and expiry of 30 days
+New key pair created with ID 114e2de892b2dcfdf96c92f3c69f39e564ecec1e and expiry of 1 day
```
marians, line 32: I think the ID in this output line should also be truncated, no? Unrelated maintenance.

axbarsan: Yes, it does have to be truncated 👍

PR #458, opened 2020-04-17
Edited for typos and English

Oshratn:


Oshratn:
I was also wondering if the reference to v.1.4.0 is redundant, or should stay on the document.

pipo02mix:
yeah now all clusters have newer versions so we can remove the reference, good point

PR #463, opened 2020-04-23
Some more adjustments for the app platform naming change.

oponder:
Hi @Oshratn , here are the other places where i thought we should adjust the language now that the concept is called App Platform.

### DIFF
--- a/src/content/reference/app-configuration/index.md
+++ b/src/content/reference/app-configuration/index.md
@@ -20,7 +20,7 @@ There are three levels of configuration:
 2. **Cluster**: Configuration provided by the cluster admin.
 3. **User**: Configuration provided by the user installing an App.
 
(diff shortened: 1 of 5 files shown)


src/content/reference/app-configuration/index.md
```diff
@@ -20,7 +20,7 @@ There are three levels of configuration:
 2. **Cluster**: Configuration provided by the cluster admin.
 3. **User**: Configuration provided by the user installing an App.
 
-Each level can overrides the previous one. As a user of the App Catalog you are not expected to edit configuration at the `catalog` or `cluster` level. However user level configuration can override both catalog and cluster level configuration.
+Each level can overrides the previous one. As a user you are not expected to edit configuration at the `catalog` or `cluster` level. However user level configuration can override both catalog and cluster level configuration.
```
Oshratn, line 23:
```suggestion
Each level can override the previous one. As a user you are not expected to edit configuration at the `catalog` or `cluster` level. However user level configuration can override both catalog and cluster level configuration.
```

PR #466, opened 2020-04-24
Fix similar instance types reference

marians:
This PR attempts to fix two problems with https://docs.giantswarm.io/reference/similar-ec2-instance-types/ :

1. The table rendering is broken
2. The page does not show up in the listing https://docs.giantswarm.io/reference/


marians:
Please ignore the broken linkcheck caused by bad URLs in the apiextension repo. Working on that.

PR #470, opened 2020-04-27
Changed headline

othylmann:
one headline seeemed wrong


oponder:
It wasn't right yet, this heading should be about the catalog that is already installed for you, not how to install it. Fixed it. Thanks for noticing something wasn't right though!

PR #474, opened 2020-04-29
Fix weirdness

webwurst:
Looks like something is missing. Not sure if I got this right, feel free to adapt.


webwurst:
Merging, failed link checks are only for example.com

PR #477, opened 2020-04-30
Add CRD metadata and reveal in index page

marians:
Towards https://github.com/giantswarm/giantswarm/issues/8804

This adds some metadata for CRDs to the [reference](https://docs.giantswarm.io/reference/cp-k8s-api/) with the purpose:

- to give users more context on the CRDs in the index
- to allow for filtering and navigation between CRDs in a next step

Also this modifies the CRD index page in these ways:

- Add the technical CRD name for more differentiation
- Add the topics and provider tags from metadata

### To reviewers

Please check the topic assignment `tenantcluster` vs. `controlplane` for the CRDs you have good knowledge about.

### Previews

image

image

### DIFF
--- a/src/data/crd_metadata.yaml
+++ b/src/data/crd_metadata.yaml
@@ -1,0 +1,103 @@
+# This data file holds meta data for CRDs
+# with the purpose of providing context and filtering capabilities for end users
+# of our CRD docs.
+#
+# Schema:
+# .topics: array of strings. Possible values: "apps", "controlplane", "tenantcluster"
+# .provider: array of strings.
+appcatalogs.application.giantswarm.io:
+  topics:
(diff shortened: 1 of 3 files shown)


src/data/crd_metadata.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - kvm
+  topics:
+    - tenantcluster
+machinedeployments.cluster.x-k8s.io:
+  provider:
+    - aws
+  topics:
+    - tenantcluster
+releases.release.giantswarm.io:
+  topics:
```
tfussell, line 98: Hard to say whether this is controlplane or tenantcluster. I guess both make sense.

PR #481, opened 2020-05-04
Tenant cluster labelling gsctl

ubergesundheit:
I want to use this PR to sketch out/document tenant cluster labelling using `gsctl`. Towards giantswarm/giantswarm#1693

Once this PR is approved, I'll start implemeting these changes in `gsctl`.

Here is a TL;DR of the changes/additions:

- `gsctl list cluster`
  - `-o json` additionally includes user defined cluster labels
  - `--selector` or `-l` can be used to filter the list of returned clusters
- `gsctl show cluster` returns user defined cluster labels for node pool clusters on AWS ([Example](https://github.com/giantswarm/docs/pull/481/commits/d801b975b5952f437b42bb640c81d1952b12d93a#diff-5eede8a45557a2cf5bd14b48d3fdc857R42-R43))
- `gsctl create cluster`
  - cluster labels can be specified through a `labels` object in the v5 yaml cluster definition
- **new**: `gsctl update cluster`
  - `--label` allows to specify a label change in form of `key=value`. Can be specified multiple times. Set to `key=` to delete a label

### DIFF
--- a/src/content/reference/gsctl/list-clusters.md
+++ b/src/content/reference/gsctl/list-clusters.md
@@ -30,12 +30,12 @@ The details displayed are:
 
 ## Argument reference
 
-- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with node pools on AWS. The default output format is `table`, which results in an output like shown above.
+- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with release version 10.0.0 and above on AWS. The default output format is `table`, which results in an output like shown above.
 - `--show-deleting`: Set this flag to also list clusters that are currently being deleted and add a `DELETING SINCE` column. Has no effect with `--output json`, as JSON output contains all clusters, deleted or not.
 - `--selector` or `-l`: Label selector query to filter clusters on.
 Accepts a single string containing multiple selector requirements which are comma-separated.
 In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.
-This feature is only available for clusters with node pools on AWS.
+This feature is only available for clusters with release version 10.0.0 and above on AWS.
 
 ## Related
 
--- a/src/content/reference/gsctl/list-clusters.md
+++ b/src/content/reference/gsctl/list-clusters.md
@@ -30,9 +30,12 @@ The details displayed are:
 
 ## Argument reference
 
-- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels. The default output format is `table`, which results in an output like shown above.
+- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with node pools on AWS. The default output format is `table`, which results in an output like shown above.
 - `--show-deleting`: Set this flag to also list clusters that are currently being deleted and add a `DELETING SINCE` column. Has no effect with `--output json`, as JSON output contains all clusters, deleted or not.
-- `--selector` or `-l`: Label selector query to filter clusters on. Accepts a single string containing multiple selector requirements which are comma-separated. In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.
+- `--selector` or `-l`: Label selector query to filter clusters on.
+Accepts a single string containing multiple selector requirements which are comma-separated.
+In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.
+This feature is only available for clusters with node pools on AWS.
 
 ## Related
 
--- a/src/content/reference/gsctl/list-clusters.md
+++ b/src/content/reference/gsctl/list-clusters.md
@@ -32,6 +32,7 @@ The details displayed are:
 
 - `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. The default output format is `table`, which results in an output like shown above.
 - `--show-deleting`: Set this flag to also list clusters that are currently being deleted and add a `DELETING SINCE` column. Has no effect with `--output json`, as JSON output contains all clusters, deleted or not.
+- `--selector` or `-l`: Label selector query to filter clusters on. Accepts a single string containing multiple selector requirements which are comma-separated. In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.
 
 ## Related
 
--- a/src/content/reference/gsctl/list-clusters.md
+++ b/src/content/reference/gsctl/list-clusters.md
@@ -28,6 +28,17 @@ The details displayed are:
 - `RELEASE`: The version number of the release used by the cluster. For older clusters, this may be `n/a`.
 - `CREATED`: The date and time when the cluster has been created (UTC).
 
+The `--selector` flag can be used to filter the output based on a set of requirements.
+
+```nohighlight
+$ gsctl list clusters --selector environment=testing
+ID     ORGANIZATION  NAME        RELEASE  CREATED
+z63so  testteam      Dan's Test  2.7.2    2017 Jun 19, 09:12 UTC
+```
+
+In this example, cluster `z63so` is the only cluster with label `environment=testing`.
+More information about possible queries can be found in the [Kubernetes Labels and Selectors documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/).
+
 ## Argument reference
 
 - `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with release version 10.0.0 and above on AWS. The default output format is `table`, which results in an output like shown above.
--- a/src/content/reference/gsctl/list-clusters.md
+++ b/src/content/reference/gsctl/list-clusters.md
@@ -30,7 +30,7 @@ The details displayed are:
 
 ## Argument reference
 
-- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. The default output format is `table`, which results in an output like shown above.
+- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains cluster labels. The default output format is `table`, which results in an output like shown above.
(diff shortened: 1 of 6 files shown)


ubergesundheit:
> I like it. The only thing I see is that the `show cluster` command indicates that this is AWS node pools only, but I don't see this mentioned in the other commands.

Good point. it is AWS node pools only. I'll add a note to the other commands

ubergesundheit:
Documentation of `gsctl update cluster --name` from giantswarm/giantswarm#7579

src/content/reference/gsctl/show-cluster.md
```diff
@@ -79,6 +81,7 @@ The output lines in detail:
 - **AWS account:** (_only on AWS_) If the cluster is running using non-default provider credentials, here we show the AWS account ID
 - **Azure subscription:** (_only on Azure_) If the cluster is running using non-default provider credentials, here we show the subscription ID
 - **Azure tenant:** (_only on Azure_) If the cluster is running using non-default provider credentials, here we show the tenant ID
+- **Labels** (_only on AWS_) user defined labels. Only available for clusters with node pools
```
marians, line 84: Actually, only on AWS for clusters using releases >v10.x.x.

Also, `Only available for clusters with node pools` sounds a bit confusing. A v11.x.x cluster can have labels even without having a node pool.

src/content/reference/gsctl/update-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+description: The 'gsctl update cluster' command allows the modification of the name of a cluster and its labels.
+date: 2020-05-04
+type: page
+weight: 44
+---
+
+# `gsctl update cluster`
+
+The `gsctl update cluster` allows the modification of the name of a cluster and its labels.
+Cluster labelling is only available for clusters with node pools on AWS.
```
marians, line 12: Actually, only on AWS for clusters using releases >v10.x.x.

src/content/reference/gsctl/update-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+### Name modification example
+
+```nohighlight
+$ gsctl update cluster f01r4 --name "Precious Production Cluster"
+```
+
+### Labels modification JSON Merge Patch example
+
+```nohighlight
+$ gsctl update cluster vxvc7 --labels '{"environment": "testing", "locked": null}'
```
marians, line 30: Not digging the syntax (use of JSON) here. Let's come up with something that is more CLI-native.

src/content/reference/gsctl/update-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+will update the labels of cluster `vxvc7`. It will add (or update depending on prior existence) label `environment=testing` and delete the label with key `locked`.
+
+### Labels modification key=value example
+
+```nohighlight
+$ gsctl update cluster vxvc7 --labels 'environment=testing locked-'
```
marians, line 38: I would strongly suggest use of multiple `--label` flags here (not `--labels`). Also I don't like the custom "minus" syntax, an empty string should be more expressive.

```suggestion
$ gsctl update cluster vxvc7 --label environment=testing --label locked=
```

In spf13/cobra, using a `StringSlice` flag for the above means that the following syntax will also work:

```
$ gsctl update cluster vxvc7 --label environment=testing,locked=
```

src/content/reference/gsctl/update-cluster.md
ubergesundheit: so only the `key=value` syntax then?

src/content/reference/gsctl/update-cluster.md
ubergesundheit: This is lifted from `kubectl labels` https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#label

src/content/reference/gsctl/show-cluster.md
```diff
@@ -39,6 +39,8 @@ Worker nodes running:          8
 Worker EC2 instance type:      m5.xlarge
 CPU cores in workers:          28
 RAM in worker nodes (GB):      26.5
+Labels:                        usage=staging
+                               locked=false
```
marians, line 43: I'd place the `Labels` row(s) under `Release version`. That way the size and resource group keeps at the end.

By the way, that example is for release version prior to v10.x.x (non node pools). For >10 it looks slightly different, as there is less worker node information.

src/content/reference/gsctl/update-cluster.md
marians: And then I found the syntax described below. When we make that usable, I see no reason to support JSON in the CLI.

src/content/reference/gsctl/update-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+will update the labels of cluster `vxvc7`. It will add (or update depending on prior existence) label `environment=testing` and delete the label with key `locked`.
+
+## Full argument reference {#arguments}
+
+- `--name` or `-n`: The new cluster name.
+- `--labels`: Specify label updates. Allows to specify label updates either as a [JSON Merge Patch, RFC 7386](https://tools.ietf.org/html/rfc7386) or in form of `'key=value anotherkey=anothervalue'`.
```
marians, line 46: Would need adaptation according to the example above

src/content/reference/gsctl/update-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+Use `key-` to delete the label with key `key`.
```
marians, line 47:
```suggestion
```

marians, line 47:
```suggestion
To remove a label, set it to an empty string.
```

src/content/reference/gsctl/update-cluster.md
ubergesundheit: What about people using gsctl in automation not wanting to talk directly to the API? I am very much fine with implementing less, but the original story (https://github.com/giantswarm/giantswarm/issues/1693) even mentions usage of tenant cluster labels in CI context

src/content/reference/gsctl/update-cluster.md
MarcelMue: I would also like to align with kubectl here as this is also going to be the interaction basis in the future while we open our control plane.

src/content/reference/gsctl/list-clusters.md
```diff
@@ -30,12 +30,12 @@ The details displayed are:
 
 ## Argument reference
 
-- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with node pools on AWS. The default output format is `table`, which results in an output like shown above.
+- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with release version 10.0.0 and above on AWS. The default output format is `table`, which results in an output like shown above.
```
marians, line 44:
```suggestion
- `--output` or `-o`: Using this flag with the value `json`, the output can be printed in JSON format. This is convenient for use in automation. JSON representation additionally contains user defined cluster labels for clusters with release version {{% first_aws_nodepools_version %}} and above on AWS. The default output format is `table`, which results in an output like shown above.
```

I just recalled that we have a shortcode to avoid duplicating this bit of information.

src/content/reference/gsctl/list-clusters.md
```diff
@@ -30,12 +30,12 @@ The details displayed are:
...
 - `--show-deleting`: Set this flag to also list clusters that are currently being deleted and add a `DELETING SINCE` column. Has no effect with `--output json`, as JSON output contains all clusters, deleted or not.
 - `--selector` or `-l`: Label selector query to filter clusters on.
 Accepts a single string containing multiple selector requirements which are comma-separated.
 In the case of multiple requirements, all must be satisfied so the comma separator acts as a logical AND (&&) operator.
-This feature is only available for clusters with node pools on AWS.
+This feature is only available for clusters with release version 10.0.0 and above on AWS.
```
marians, line 49:
```suggestion
This feature is only available for clusters with release version {{% first_aws_nodepools_version %}} and above on AWS.
```

src/content/reference/gsctl/show-cluster.md
```diff
@@ -81,7 +81,7 @@ The output lines in detail:
 - **AWS account:** (_only on AWS_) If the cluster is running using non-default provider credentials, here we show the AWS account ID
 - **Azure subscription:** (_only on Azure_) If the cluster is running using non-default provider credentials, here we show the subscription ID
 - **Azure tenant:** (_only on Azure_) If the cluster is running using non-default provider credentials, here we show the tenant ID
-- **Labels** (_only on AWS_) user defined labels. Only available for clusters with node pools
+- **Labels** (_only on AWS_) user defined labels. Only available for clusters with release version 10.0.0 and above on AWS
```
marians, line 109:
```suggestion
- **Labels** (_only on AWS_) user defined labels. Only available for clusters with release version {{% first_aws_nodepools_version %}} and above on AWS
```

src/content/reference/gsctl/update-cluster.md
```diff
@@ -9,7 +9,7 @@ weight: 44
 # `gsctl update cluster`
 
 The `gsctl update cluster` allows the modification of the name of a cluster and its labels.
-Cluster labelling is only available for clusters with node pools on AWS.
+Cluster labelling is only available for clusters with release version 10.0.0 and above on AWS.
```
marians, line 12:
```suggestion
Cluster labelling is only available for clusters with release version {{% first_aws_nodepools_version %}} and above on AWS.
```

src/content/reference/gsctl/update-cluster.md
marians: > What about people using gsctl in automation not wanting to talk directly to the API?

They should be able to use `--label` one or several times in the non-JSON way.

src/content/reference/gsctl/update-cluster.md
marians: To me, aligning with a very specific kubectl behaviour in a tool that is in many ways different from kubectl is not a goal we should pursue.

That would be different if gsctl was a kubectl plugin already.

src/content/reference/gsctl/list-clusters.md
ubergesundheit: Very nice! Adding this :+1:

src/content/reference/gsctl/show-cluster.md
```diff
@@ -41,6 +41,33 @@ CPU cores in workers:          28
 RAM in worker nodes (GB):      26.5
 ```
 
+Example output for an AWS based cluster with release version 10.0.0 and above:
```
ubergesundheit, line 44:
```suggestion
Example output for an AWS based cluster with release version {{% first_aws_nodepools_version %}} and above:
```

src/content/reference/gsctl/update-cluster.md
ubergesundheit: We'll do it the multiple `--label` way to get implementation tires kicking..

PR #493, opened 2020-05-12
Update coredns docs for the app cr migration.

oponder:
This updates coredns docs to explain where the user values configmap is (and the format) depending on the release version.

Would appreciate a fact check from anyone that knows the specifics. ping @omegas27  @rossf7 @pipo02mix

Towards: giantswarm/giantswarm#10211

### DIFF
--- a/src/content/guides/advanced-coredns-configuration/index.md
+++ b/src/content/guides/advanced-coredns-configuration/index.md
@@ -9,27 +9,105 @@ tags: ["tutorial"]
 
 # Advanced CoreDNS Configuration
 
-The [CoreDNS addon](https://github.com/coredns/coredns) running inside your cluster has additional configuration options and features that can be customized.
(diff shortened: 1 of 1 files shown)


src/content/guides/advanced-coredns-configuration/index.md
```diff
@@ -9,27 +9,105 @@ tags: ["tutorial"]
...
 
-__Warning:__ Please do not edit any of the other CoreDNS related resources. Only the user ConfigMap is safe to edit.
+__Warning:__
+
+Please do not edit any other coredns related ConfigMaps.
+
+Only the user values ConfigMap is safe to edit.
+
+------
+
```
pipo02mix, line 54: Is this right? It looks like broken in the preview

oponder: hmm, looks how i intended it to, which is to give it a line above and below, for extra visibility

PR #497, opened 2020-05-13
Paragraph line-height and heading changes.

oponder:
I found the typography of the docs a bit cramped in some way. 

Also adjusted the margins around headings to try to make it clearer to which section it belongs and to have sections appear more clearly in general.

Also removed the different styling, colors, and font-weights of some of the smaller headings.

changes

### DIFF
--- a/src/static/css/base.sass
+++ b/src/static/css/base.sass
@@ -6,34 +6,39 @@
 
 h1, h2, h3, h4, h5, h6
   text-align: left
+  line-height: 1.3em
+  margin-bottom: 15px
+
+h1
+  font-size: 42px
+
 
 h2
-  font-size: 36px
+  font-size: 33px
   font-weight: 300
-  margin-top: 35px
-  margin-bottom: 30px
+  margin-top: 45px
+  line-height: 1.45em
 
 h3
-  font-size: 28px
-  margin-top: 20px
-  margin-bottom: 20px
-  color: #666
+  font-size: 26px
+  margin-top: 45px
 
 h4
-  font-weight: 700
-  color: #888
-  font-size: 18px
-  margin-top: 20px
-  margin-bottom: 20px
+  font-size: 20px
+  margin-top: 45px
 
 h5
   font-weight: 700
   font-size: 16px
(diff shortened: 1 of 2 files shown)


src/static/css/base.sass
```diff
@@ -6,34 +6,39 @@
...
+
 
 h2
-  font-size: 36px
+  font-size: 33px
   font-weight: 300
-  margin-top: 35px
-  margin-bottom: 30px
+  margin-top: 45px
+  line-height: 1.45em
```
axbarsan, line 20: Why not have line heights directly related to the element's font size, rather than related to the root font size? It would be much easier to adjust in the future.

You can do that by setting a number to the `line-height` property, e. g.:

```css
font-size: 16px;

/* This will have the same value as the font size = 16px */
line-height: 1; 

/* This will have the font size * 1.2 = 19,2px */
line-height: 1.2; 
```

oponder: 🤔 

I thought thats what `em` did?  https://www.w3schools.com/cssref/css_units.asp

> em: Relative to the font-size of the element (2em means 2 times the size of the current font)




For root size it would be `rem`. 

Not sure what happens when you leave out the unit altogether, maybe it indeed defaults to `em`?

axbarsan: Yeah, you’re right. I was mistaking `em` for `rem` 🤦‍♂️
For line-heights, leaving out the unit is the same as using `rem`

oponder:
pinging @marians also for typography critique :)

marians:
Thanks for the ping! From the screenshot, I like the headline change, but am thinking that the space before and after headlines needs to be adjusted (increased) accordingly. Not sure yet about the line height in normal text. Will run a preview later.

oponder:
@marians feel free to check it out and commit your changes if you find them

I mostly wanted to have more space for the content to breath and for different sections to become more apparent, so if that's the direction you're going in, im cool with that.

oponder:
Gonna go ahead and merge this, feel free to refine in subsequent PR's

PR #516, opened 2020-05-27
Move BYOC images to multi-account folder

oponder:
FYI @oshrat, this moves the images to the new folder and deletes the BYOC folder


Oshratn:
Thanks 🌹

PR #534, opened 2020-06-17
Adjust list item line-height and margin-bottom

marians:
This adjust the `<li>` appearance to a normal paragraph.

## Before

image

## After

image

### DIFF
--- a/src/static/css/base.sass
+++ b/src/static/css/base.sass
@@ -35,6 +35,10 @@ h5
 p
   line-height: 1.65em
 
+li
+  line-height: 1.65em
+  margin-bottom: 11px
+
 pre
   background-color: #002b36
   padding: 4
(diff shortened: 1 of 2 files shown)


src/static/css/base.sass
```diff
@@ -35,6 +35,10 @@ h5
 p
   line-height: 1.65em
 
+li
+  line-height: 1.65em
+  margin-bottom: 11px
```
axbarsan, line 40: Do we want all list items to have a bottom margin, or do we just want to set the spacing between list items?

marians: Actually I think between would be preferrable. But on the other hand, we don't bother with `<p>` either. Maybe we can fix it for both.

axbarsan: Alright, we can live that for another PR 👍

PR #539, opened 2020-06-19
Update guide on cluster creation via CP API for HA masters

marians:
For https://github.com/giantswarm/giantswarm/issues/8346

- Adds information on the G8sControlPlane and AWSControlPlane resources
- Improves language and readability
- Fixes a CR example where we showed an AWSMachineDeployment instead of AWSCluster
- Fixes headline levels in HA masters intro page

### DIFF
--- a/src/content/guides/creating-clusters-via-crs-on-aws/index.md
+++ b/src/content/guides/creating-clusters-via-crs-on-aws/index.md
@@ -6,33 +6,38 @@ type: page
 weight: 100
 tags: ["tutorial"]
 ---
+
 # Creating tenant clusters via the Control Plane Kubernetes API
 
-This guide will show you how to create tenant clusters by creating and applying CRs directly to the control plane.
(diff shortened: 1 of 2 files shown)


src/content/guides/creating-clusters-via-crs-on-aws/index.md
```diff
@@ -54,39 +59,95 @@ spec:
...
-    availabilityZones:
-    - eu-central-1a
-    worker:
-      instanceType: m4.xlarge
+    credentialSecret:
+      name: credential-default
+      namespace: giantswarm
+    master:
+      availabilityZone: ""
+      instanceType: ""
```
xh3b4sd, line 89: We need to migrate this structure here away in the mid term. Should we still show it in the docs?

marians: If the CR is valid without the `master` part I am all for removing it in docs.

xh3b4sd: Looks pretty optional to me: https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/infrastructure/v1alpha2/aws_cluster_types.go#L116-L119.

PR #544, opened 2020-07-01
Update node pools information regarding spot instances

marians:
Towards https://github.com/giantswarm/giantswarm/issues/11896

Changes:

- Make the spot instances topic easier to discover by changing the sub heading from "Instance distribution" to "On-demand and spot instances"
- Add general explanation of what spot differentiates form on-demand
- Clarify that there is not automatic replacement of missing spot instances through on-demand instances
- Add hints on how to increase the likelihood of getting spot isntances

### DIFF
--- a/src/content/basics/nodepools/index.md
+++ b/src/content/basics/nodepools/index.md
@@ -1,7 +1,7 @@
 ---
 title: Node Pools
 description: A general description of node pools as a concept, it's benefits, and some details you should be aware of.
-date: 2020-04-28
+date: 2020-07-01
 weight: 130
 type: page
 categories: ["basics"]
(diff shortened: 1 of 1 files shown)


xh3b4sd:
> Clarify that there is not automatic replacement of missing spot instances through on-demand instances

Off topic. This sounds like a super useful feature we should provide. Is this possible out of the box with the AWS primitives? Do we plan on pursuing this?

src/content/basics/nodepools/index.md
```diff
@@ -112,17 +112,27 @@ Pay close attention to the workloads being rescheduled on other nodes once nodes
 
 See the [`gsctl delete nodepool`](/reference/gsctl/delete-nodepool/) reference for how to delete a node pool using the CLI.
 
-## Instance distribution
+## On-demand and spot instances {#on-demand-spot}
```
xh3b4sd, line 115: I myself failed to find the docs in the past because of that. There was just no real search result for "spot instances".

marians:
> Off topic. This sounds like a super useful feature we should provide. Is this possible out of the box with the AWS primitives? Do we plan on pursuing this?

Yes. There is method where you set up two node pools, one with spot, one with on-demand. Then you configure cluster-autoscaler to only scale up on-demand when spot cannot scale. See

https://blog.doit-intl.com/running-eks-workloads-on-spot-instances-with-on-demand-instances-fallback-14bef39ce689

I would like to ensure that it works for us and then create a guide and link to it from this page.

PR #548, opened 2020-07-06
Remove dead links

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/11852

The links died as a result of k8scc flattening. Current content is [here](https://github.com/giantswarm/k8scloudconfig/tree/master/files/k8s-resource) but I'm not sure the links add that much value and they will die again once ignition-operator is implemented. This content seems to still be up to date otherwise, but LMK if not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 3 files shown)


stone-z:
CI fails due to old architect-orb. Might as well bump that while I'm here

PR #551, opened 2020-07-16
Add missing CRD meta data

marians:
This adds topic and provider metadata for newer CRDs.

As a result, entries on https://docs.giantswarm.io/reference/cp-k8s-api/ will show up with topic and provider labels.


marians:
Since the new CRDs are mostly or all related to Azure, pinging team celestial for review here. Please check whether the right tenantcluster/controlplane topics are applied and whether the `azure` provider appears where it should appear. Thanks!

PR #558, opened 2020-08-09
Document multi-nginx support

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/9060

### DIFF
--- a/src/content/guides/multi-nginx/index.md
+++ b/src/content/guides/multi-nginx/index.md
@@ -8,7 +8,7 @@ tags: ["tutorial"]
 
 # Multi-NGINX ingress controller support
 
-NGINX ingress controller handles Ingress resources, routing the traffic from outside of the Kubernetes cluster to services within the cluster. Starting with App version 1.8.0 one can install multiple NGINX ingress controllers in same Kubernetes cluster. Here we cover common cases where having multiple NGINX ingress controllers can be useful.
+NGINX ingress controller handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources, routing the traffic from outside of the Kubernetes cluster to services within the cluster. Starting with App version 1.8.0 one can install multiple NGINX ingress controllers in same Kubernetes cluster. Here we cover common cases where having multiple NGINX ingress controllers can be useful.
 
 Most of the NGINX configuration options have NGINX wide defaults but can also be overriden on per Ingress resource level. In each of the covered multi-NGINX use cases, one essentially installs second NGINX with different NGINX global-only configuration, i.e. configuration that applies to all Ingress resources managed by that NGINX installation which cannot be customized on per-Ingress resource level.
 
@@ -75,3 +75,7 @@ Here is how this can be achieved:
   ```
 
 For the second NGINX IC App installation ingress class and host name subdomain are customized for uniqueness. Additionally, default SSL ciphers are overriden to include weak ciphers for legacy clients.
+
+## Further reading
+
+- [Official ingress-nginx multi-nginx documentation](https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/)
--- a/src/content/guides/multi-nginx/index.md
+++ b/src/content/guides/multi-nginx/index.md
@@ -6,7 +6,7 @@ type: page
 tags: ["tutorial"]
 ---
 
-# Multi-NGINX ingress controller support
+# Running Multiple NGINX Ingress Controllers
 
 NGINX ingress controller handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources, routing the traffic from outside of the Kubernetes cluster to services within the cluster. Starting with App version 1.8.0 one can install multiple NGINX ingress controllers in same Kubernetes cluster. Here we cover common cases where having multiple NGINX ingress controllers can be useful.
 
--- a/src/content/guides/multi-nginx/index.md
+++ b/src/content/guides/multi-nginx/index.md
@@ -58,7 +58,7 @@ In NGINX IC App v1.2.0 upgrade to ingress-nginx 0.27.1 was included. Among other
 
 With single NGINX one had option of restoring weak SSL ciphers configuration, to support few services with older clients until the clients get upgraded. Problem with this approach, since SSL ciphers are global settings, was that changing default SSL ciphers back by restoring weak ciphers would apply to all Ingresses and service behind them, not just the one with old clients.
 
-With multiple NGINX ingress controller support, one can have separate NGINX installations with different SSL ciphers configuration, to localize but still make accessible services used by the legacy clients.
+With multiple NGINX ingress controller support, one can have separate NGINX installations with different SSL ciphers configuration, to localize but still make services used by legacy clients accessible.
 
 Here is how this can be achieved:
 
--- a/src/content/guides/multi-nginx/index.md
+++ b/src/content/guides/multi-nginx/index.md
@@ -1,0 +1,77 @@
+---
+title: Multi-NGINX ingress controller support
+description: Deploy multiple NGINX ingress controllers in a Kubernetes cluster to separate different ingress traffic classes.
+date: 2020-08-05
+type: page
+tags: ["tutorial"]
+---
+
+# Multi-NGINX ingress controller support
+
+NGINX ingress controller handles Ingress resources, routing the traffic from outside of the Kubernetes cluster to services within the cluster. Starting with App version 1.8.0 one can install multiple NGINX ingress controllers in same Kubernetes cluster. Here we cover common cases where having multiple NGINX ingress controllers can be useful.
+
+Most of the NGINX configuration options have NGINX wide defaults but can also be overriden on per Ingress resource level. In each of the covered multi-NGINX use cases, one essentially installs second NGINX with different NGINX global-only configuration, i.e. configuration that applies to all Ingress resources managed by that NGINX installation which cannot be customized on per-Ingress resource level.
+
+## Separating public from internal ingress traffic
+
+Having some components in multi-tierd architecture or back-office applications not accessible via public internet is a common requirement for security or performance reasons. This is how one can achieve it by using multi-NGINX ingress controller support:
+
+- Deploy one NGINX IC App using default settings, for making selected cluster services accessible via the public internet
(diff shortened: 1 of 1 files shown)


sslavic:
Thanks @webwurst for review!

multi-nginx is not supported yet in upstream chart, so that is not covered in the upstream docs, there's nothing to link to.

It is supported in upstream binary and I've added a link to the upstream docs https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/ It's rather low level stuff that we made use of in the chart.

cokiengchiara:
Thanks so much for writing this @sslavic . It flows well, covers the most important use cases, and very actionable.

High level questions:
* If one wants to have multiple NGINX for different reasons, is it clear which values can be overridden? 
* The other two use cases we talked about was supporting (1) dev vs prod traffic and (2) help w hype traffic - would it be useful to cover that here? Or is that obvious?

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Multi-NGINX ingress controller support
```
cokiengchiara, line 2:
```suggestion
title: Running Multiple NGINX Ingress Controllers
```

Does this change make sense to make it more searchable?

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
...
+date: 2020-08-05
+type: page
+tags: ["tutorial"]
+---
+
+# Multi-NGINX ingress controller support
+
+NGINX ingress controller handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources, routing the traffic from outside of the Kubernetes cluster to services within the cluster. Starting with App version 1.8.0 one can install multiple NGINX ingress controllers in same Kubernetes cluster. Here we cover common cases where having multiple NGINX ingress controllers can be useful.
+
+Most of the NGINX configuration options have NGINX wide defaults but can also be overriden on per Ingress resource level. In each of the covered multi-NGINX use cases, one essentially installs second NGINX with different NGINX global-only configuration, i.e. configuration that applies to all Ingress resources managed by that NGINX installation which cannot be customized on per-Ingress resource level.
```
cokiengchiara, line 13: Most of the NGINX configuration options have NGINX wide defaults but can also be overriden on per Ingress resource level. 
* Would it be helpful to specify what, in general, **can** or **cannot** be overridden? (whatever is fewer)

In each of the covered multi-NGINX use cases, one essentially installs second NGINX with different NGINX global-only configuration, i.e. configuration that applies to all Ingress resources managed by that NGINX installation which cannot be customized on per-Ingress resource level.
* "different NGINX global-only configuration" was confusing to me. But I don't know what to suggest. Is "global-only configuration" a specific term?
* Would sth like, "i.e. configurations managed by an NGINX installation that cannot be customized per-Ingress." be shorter, but clearer?

In each of the covered multi-NGINX use cases, one essentially installs second NGINX with different global NGINX-only configuration, i.e. configuration that applies to all Ingress resources managed by that NGINX installation which cannot be customized on per-Ingress resource level.

cokiengchiara:
> * (1) dev vs prod traffic

I think this is already the external vs internal right.

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Separating public from internal ingress traffic
+
+Having some components in multi-tierd architecture or back-office applications not accessible via public internet is a common requirement for security or performance reasons. This is how one can achieve it by using multi-NGINX ingress controller support:
+
+- Deploy one NGINX IC App using default settings, for making selected cluster services accessible via the public internet
```
cokiengchiara, line 19: Would it be helpful to link to docs here on default settings?

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
...
+        # enabled: true
+        public: false
+        subdomain: "ingress-internal"
+        # type: NodePort
+        nodePorts:
+          http: 31010
+          https: 31011
+    ```
+
+Each NGINX IC App installation has to have unique ingress class. Ingress resources can then declare which ingress controller should be handling their route definition by referencing appropriate ingress class in `kubernetes.io/ingress.class` annotation. Default NGINX IC App ingress class is `nginx`. In the above example we configure `nginx-internal` as the second NGINX IC App installation's ingress class.
```
cokiengchiara, line 47:
```suggestion
Each NGINX IC App installation has to have unique `ingressClass`. Ingress resources can then declare which ingress controller should be handling their route definition by referencing appropriate `ingressClass` in `kubernetes.io/ingress.class` annotation. Default NGINX IC App ingress class is `nginx`. In the above example we configure `nginx-internal` as the second NGINX IC App installation's ingress class.
```

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Using weak ciphers for legacy clients
+
+In NGINX IC App v1.2.0 upgrade to ingress-nginx 0.27.1 was included. Among other things this upgrade brought in security improvements - SSL ciphers which are considered as weak got removed from the default configuration. Some of the older clients (like web browsers, http libraries in apps) could no longer establish secure connections with cluster services exposed via new NGINX, since these legacy clients only supported SSL ciphers that got removed.
+
+With single NGINX one had option of restoring weak SSL ciphers configuration, to support few services with older clients until the clients get upgraded. Problem with this approach, since SSL ciphers are global settings, was that changing default SSL ciphers back by restoring weak ciphers would apply to all Ingresses and service behind them, not just the one with old clients.
+
+With multiple NGINX ingress controller support, one can have separate NGINX installations with different SSL ciphers configuration, to localize but still make accessible services used by the legacy clients.
```
cokiengchiara, line 61:
```suggestion
With multiple NGINX ingress controller support, one can have separate NGINX installations with different SSL ciphers configuration, to localize but still make services used by legacy clients accessible.
```

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
...
+  ```yaml
+  controller:
+    ingressClass: "nginx-weak"
+    service:
+      subdomain: "ingress-weak"
+  configmap:
+    ssl-ciphers: "ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256"
+  ```
+
+For the second NGINX IC App installation ingress class and host name subdomain are customized for uniqueness. Additionally, default SSL ciphers are overriden to include weak ciphers for legacy clients.
```
cokiengchiara, line 77: is the syntax or format for ssl-ciphers clear? (I am just asking)

src/content/guides/multi-nginx/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Running Multiple NGINX Ingress Controllers
+description: Deploy multiple NGINX ingress controllers in a Kubernetes cluster to separate different ingress traffic classes.
+date: 2020-08-05
+type: page
+tags: ["tutorial"]
+---
+
+# Multi-NGINX ingress controller support
```
cokiengchiara, line 9:
```suggestion
# Running Multiple NGINX Ingress Controllers
```

sslavic:
> * If one wants to have multiple NGINX for different reasons, is it clear which values can be overridden?

- all can; just because it can doesn't mean it should be overridden; PR already covers that it makes sense to have separate nginx for different global-only settings; even for ones that can be overridden on per Ingress level it may make sense to have separate nginx instead; it's all context dependent; if/when we get support requests or learn more ourselves, we can improve docs further with additional use cases

> * The other two use cases we talked about was supporting (1) dev vs prod traffic and (2) help w hype traffic - would it be useful to cover that here? Or is that obvious?

- (1) dev vs prod is only theoretical use case - nobody sane uses single k8s cluster for different environments
- (2) I'm not convinced, have to think more if it would help (so I'm reluctant to promote) to have (just) separate nginx for flash sale related services/apps and separate for the rest; one can achieve this kind of separation and more by having multiple k8s clusters

src/content/guides/multi-nginx/index.md
sslavic: > Would it be helpful to specify what, in general, can or cannot be overridden? (whatever is fewer)

No, there's extensive upstream nginx configuration docs, and they keep on changing. I've added a link, to both upstream and our chart configuration options.


> "different NGINX global-only configuration" was confusing to me. But I don't know what to suggest. Is "global-only configuration" a specific term?

I made that one up, there's no official term.

> Would sth like, "i.e. configurations managed by an NGINX installation that cannot be customized per-Ingress." be shorter, but clearer?

Maybe it's better, but it's not shorter compared to "NGINX global-only configuration". I've expanded what is meant under that term in the continuation of the sentence.

src/content/guides/multi-nginx/index.md
sslavic: I've added link to our app/chart configuration options.

src/content/guides/multi-nginx/index.md
sslavic: Ingress class is configured via `controller.ingressClass` config property in NGINX, but via `kubernetes.io/ingress.class` annotation in Ingress resources.

I was deliberately not referring to either of the two but to ingress class concept as such.

Things are evolving, in new k8s 1.18 https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class
instead of annotation a IngressClass/ingressClassName got specified for use in Ingress resources.

Whatever the means are to specify ingress class (old or new), each NGINX will have to have a unique one associated with it, and Ingress resources will have to have be classified with one of them.

src/content/guides/multi-nginx/index.md
sslavic: For the curious ones, upstream docs cover the syntax.

PR #561, opened 2020-08-25
azure v5: modify gsctl nodepools-related documentation to mention Azure support

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12706

This PR adds Azure-specific gsctl nodepools documentation.


src/layouts/shortcodes/first_azure_nodepools_version.html
```diff
@@ -0,0 +1 @@
+12.1.0
```
axbarsan, line 1: This is so we could use `{{% first_azure_nodepools_version %}}` to get first Azure NP release.

tuommaki (approved):
LGTM

PR #567, opened 2020-09-16
Routine external links check for docs

marians:
This PR contains a dummy change to be able to trigger a check of external links.


marians:
One error found in an external link, which wasn't reproducible.

URL https://sysdig.com/blog/container-isolation-gone-wrong/ works for me.

PR #570, opened 2020-09-22
Update ingress api versions and annotation for nginx ingress controller

ubergesundheit:
This PR updates the api versions of examples regarding ingress and changes an annotation.

Related: https://gigantic.slack.com/archives/C13MG9F0W/p1600735827002400

### DIFF
--- a/src/content/guides/accessing-services-from-the-outside/index.md
+++ b/src/content/guides/accessing-services-from-the-outside/index.md
@@ -62,7 +62,7 @@ You can expose Services publicly by setting up a simple Ingress. You can do this
 Let's look at the following YAML, which you can use as a template for your Ingress manifest:
 
 ```yaml
-apiVersion: extensions/v1beta1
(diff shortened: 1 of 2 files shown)


MarcelMue (approved):
LGTM from the report on slack. I am no Ingress wizard though.

tfussell:
I apologize @ubergesundheit. When I was looking at it yesterday it seemed like it had been `v1` for a while, but now I see that it was only promoted to GA in 1.19. We should still update the docs, but we need to get 1.19 released on our platforms first. I think the other change is useful and correct. Maybe we could add a note about the pre-1.19 and post-1.19 method or just wait until 1.19 is ready to merge this.

ubergesundheit:
Switched to `networking.k8s.io/v1beta1` which seems to be the version available until 1.22

tfussell (approved):
Sorry for the trouble. LGTM.

PR #572, opened 2020-09-23
azure v5: mention Azure support in cluster definition

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12706

* Mention support for Azure Node Pools
* Change first Azure Node Pools to `v12.2.0`
* Fix indentation in the V5 node pool keys, since [they were not displayed properly nested](https://docs.giantswarm.io/reference/cluster-definition/#v5-nodepool-keys)
* Configure linter rule for lists to require 4 spaces for each level, since that's how Hugo is configured to work


MartaD (approved):
LGTM! 🚀

PR #574, opened 2020-09-23
Update apiextensions commit reference to latest release

tuommaki:
Latest apiextensions release drops CRDs for managed AKS tenant clusters
which is not a thing we use nor support.


MartaD:
Thank you! 🚀

PR #575, opened 2020-09-23
Upgrade dependency versions

ubergesundheit:
- Upgrades hugo to 0.75.1 extended
- Upgrades architect-orb to 0.11.0
- Upgrades nginx to 1.18
- Enable gzipped responses
- Fiddle around with Makefile for updates mentioned above


marians:
That's a lot to test. Please give me a few days. 🥵

marians:
A quick comment: we have an nginx in front of docs (https://github.com/giantswarm/docs-proxy). Not sure you were fully aware. gzip should be handled there.

ubergesundheit:
> A quick comment: we have an nginx in front of docs (https://github.com/giantswarm/docs-proxy). Not sure you were fully aware. gzip should be handled there.

Nginx is able to serve precompressed files directly. True, it is possible to gzip responses on the fly for proxied requests, but I think the server serving the files directly should also be responsible for serving the compressed content.

marians:
FYI: Just ran `make` without reading anything beforehand. Got this error:

    Error: Error building site: "/docs/content/guides/kubernetes-gpu/index.md:1:1": plain HTML documents not supported

The source page is in the external repo here: https://github.com/giantswarm/kubernetes-gpu/blob/master/docs/index.md

```
$ cat build/content/guides/kubernetes-gpu/index.md
<hr>
You can collaborate on this recipe on [GitHub](https://github.com/giantswarm/kubernetes-gpu).
```

However when I removed the `vendor` and `build` folders and tried again, the error did not occur.

Dockerfile
```diff
@@ -1,15 +1,26 @@
...
-WORKDIR /docs/build
-ADD ./build /docs/build
-RUN /usr/bin/hugo version
-RUN /usr/bin/hugo --destination /usr/share/nginx/html
+COPY build .
 
-USER 101
+RUN hugo --verbose --gc --minify --cleanDestinationDir --path-warnings --destination /public
+
+RUN find /public \
```
marians, line 11: A comment above this one would help

Dockerfile
```diff
@@ -1,15 +1,26 @@
...
+  -size +512c \
+  -iregex '.*\.(css|csv|html?|js|svg|txt|xml|json|webmanifest|ttf)' \
+  -exec gzip -9 -k '{}' \;
+
+FROM quay.io/giantswarm/nginx:1.18-alpine
+
+COPY nginx.conf /etc/nginx/nginx.conf
+
+RUN rm /docker-entrypoint.d/* && \
+  sed -i -e 's|listen       80;|listen 8080;|' -e 's|listen  [::]:80;||' /etc/nginx/conf.d/default.conf
```
marians, line 22: Is this doing multiple replacements? If yes, I would prefer two separate sed calls in separate lines, for readability.

Makefile
```diff
@@ -7,26 +7,13 @@ CRD_DOCS_GENERATOR_VERSION=0.1.1
 
 default: docker-build
 
-build-css:
-	docker run -it \
-		-v ${PWD}/src/static/css:/sass \
-		ellerbrock/alpine-sass /sass/base.sass /sass/base.css -m auto
-
```
marians, line 14: Nice to get rid of sass as an extra dependency! Is this handled by HUGO directly?

marians (approved):
Huge effort, and everything seems to work as expected. ❤️

Makefile
ubergesundheit: Yes. The "extended" edition of hugo includes sass compilation. :tada:

marians:
@ubergesundheit Is this ready to be finished off?

ubergesundheit:
:sweat_smile: Honestly I thought I already had merged this

PR #584, opened 2020-09-25
Add hyperlinks section to docs editors guide

marians:


cokiengchiara (approved):
Thanks, I needed this 🙈

axbarsan (approved):

ubergesundheit:
IMO it would be even better to use hugos `ref` and `relref` shortcodes (https://gohugo.io/content-management/cross-references/). It allows hugo to check the links defined at build time. 

I've also seen #583 If you think this would be a good thing to have I could issue another PR for this.

marians:
I didn't know about the ref and relref shortcodces until now. Looking at it, it seems like an additional burden to content editors. Since we want to keep the threshold to editing low, I would not recommend using them.

We have internal link integrity checking in CI, so we wouldn't gain a lot with the shortcodes IMO.

PR #585, opened 2020-09-29
Drop "Policy" suffix from names ; Drop ARN part

kopiczko:
- "GiantSwarmAWSOperator" is the policy name we currently have in all
  our CP accounts.
- ARNs are predictable - no need to copy them. The important bit is that
  we have roles named consistently.


corest (approved):

marians:
Can't judge here. Deferring to @paurosello .

marians:
@paurosello Is this fine with you?

paurosello (approved):
Yes, the terraform modules were updated and customer notifications have been sent already :)

kopiczko:
Going with that. There will be another PR comming because now we setup CP/TC accounts the same way.

PR #592, opened 2020-10-05
use some nginx directives from base container image

ubergesundheit:
This PR uses some nginx directives from the underlying `nginxinc/nginx-unprivileged:1.18-alpine` image. This should fix the permission problems.

### DIFF
--- a/helm/docs-app/templates/deployment.yaml
+++ b/helm/docs-app/templates/deployment.yaml
@@ -34,9 +34,6 @@ spec:
             limits:
               cpu: 500m
               memory: 20M
-          volumeMounts:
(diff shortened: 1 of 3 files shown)


marians:
Can you check whether we are still using the cache dir defined in https://github.com/giantswarm/docs/blob/master/helm/docs-app/templates/deployment.yaml#L37-L39 ?

marians (approved):

ubergesundheit:
Looks like the directory is present already in the base container image. Running the docs image locally does not seem to do anything in there. I'll remove it for now. We can re-add later if needed

PR #593, opened 2020-10-05
Fix CRD reference

marians:
This PR solves the problem that CRD docs appears mostly empty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ COMPANY=giantswarm
 REGISTRY=quay.io
 SHELL=bash
 MARKDOWNLINT_IMAGE=06kellyjac/markdownlint-cli:0.21.0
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@ COMPANY=giantswarm
 REGISTRY=quay.io
 SHELL=bash
 MARKDOWNLINT_IMAGE=06kellyjac/markdownlint-cli:0.21.0
-CRD_DOCS_GENERATOR_VERSION=0.1.1
+CRD_DOCS_GENERATOR_VERSION=0.1.2
```
marians, line 6: This update fixes a whitespace problem which would result in the `Required` label appear in it's own paragraph.

ubergesundheit (approved):

PR #598, opened 2020-10-07
Remove the mentions of Azure supporting any kind of autoscaling

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13649


whites11:
13.0.0 will have autoscaling AFAWK.
I would leave this docs untouched.
WDYT?

axbarsan:
> 13.0.0 will have autoscaling AFAWK.
> I would leave this docs untouched.
> WDYT?

So we'll have it the final version, but not in the `alpha`? Are we sure about that?

fiunchinho:
The reasoning is that the final version will use k8s 1.18, which is supposed to fix the issues with the autoscaler. Also, the final version will be tested before releasing.

axbarsan:
Closing as per the comments above. Moved the cluster autoscaler docs removal to https://github.com/giantswarm/docs/pull/599

PR #601, opened 2020-10-14
Remove highlight.js from docs, use server-side syntax highlighting

marians:
Closes https://github.com/giantswarm/giantswarm/issues/13766

This removes highlight.js and activates server-side syntax highlighting.

Note: Although both are named `solarized-dark`, the theme colors differ between Chroma (used by HUGO) and highlight.js.

### Before (highlight.js)

image

### After (Chroma)

image


ubergesundheit (approved):
This is nice! Less downloaded resources :+1:

PR #606, opened 2020-10-21
Modify upgrades doc

cokiengchiara:
Saw in [support](https://gigantic.slack.com/archives/C6L8J93N0/p1603261460017400) from @paurosello that nodes are drained and recreated at the same time. Updating doc to reflect this. cc @giantswarm/sig-docs

### DIFF
--- a/src/content/reference/cluster-upgrades/_index.md
+++ b/src/content/reference/cluster-upgrades/_index.md
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
 
 In particular this means:
 
-- Nodes will be drained, then stopped, then recreated.
+- Nodes are drained and recreated at the same time.
 - As a consequence of draining, Pods running on a node will be rescheduled to other nodes.
 - Once the master node is taken down and recreated, the Kubernetes API will be unavailable for a short time.
 
@@ -120,7 +120,7 @@ After recreation, worker nodes are **not expected to have the same names** they 
 In process of the worker node recreation, any data stored in worker node's local storage, i. e. outside of EBS volumes, is lost.
 
 **Note:** We are in the process of making upgrades on AWS less disruptive.
-One goal is to make new nodes available first, then drain and remove old nodes. Also we are aware that for larger clusters, one thirds of worker nodes becoming unavailable at the same time is often too much.
+For example, we are aware that for larger clusters, one thirds of worker nodes becoming unavailable at the same time is often too much.
 
 ### Provider-specific details for Azure
 
--- a/src/content/reference/cluster-upgrades/_index.md
+++ b/src/content/reference/cluster-upgrades/_index.md
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
 
 In particular this means:
 
-Nodes will be drained, then stopped, then recreated.
+- Nodes will be drained, then stopped, then recreated.
 - As a consequence of draining, Pods running on a node will be rescheduled to other nodes.
 - Once the master node is taken down and recreated, the Kubernetes API will be unavailable for a short time.
 
--- a/src/content/reference/cluster-upgrades/_index.md
+++ b/src/content/reference/cluster-upgrades/_index.md
@@ -107,7 +107,7 @@ In particular this means:
 - As a consequence of draining, Pods running on a node will be rescheduled to other nodes.
 - Once the master node is taken down and recreated, the Kubernetes API will be unavailable for a short time.
 
-**Note**: By default, tenant clusters have one master node each. Consider using High Availability masters setup to avoid API downtime during upgrades. (See: [High availability Kubernetes masters](https://docs.giantswarm.io/basics/ha-masters/))
+**Note**: By default, tenant clusters have one master node each. [Consider using a High Availability masters setup](#checklist-ha-masters) to avoid API downtime during upgrades.
 
 ### Provider-specific details for AWS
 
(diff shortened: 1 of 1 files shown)


src/content/reference/cluster-upgrades/_index.md
```diff
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
 
 In particular this means:
 
-- Nodes will be drained, then stopped, then recreated.
+- Nodes are drained and recreated at the same time.
 - As a consequence of draining, Pods running on a node will be rescheduled to other nodes.
```
cokiengchiara, line 107: Based on nodes now being drained and recreated at the same time, does this still apply?

src/content/reference/cluster-upgrades/_index.md
```diff
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
...
 - Once the master node is taken down and recreated, the Kubernetes API will be unavailable for a short time.
```
cokiengchiara, line 108: Based on nodes now being drained and recreated at the same time, does this still apply?

cokiengchiara:
Also, is this still true?

> **Note**: Currently tenant clusters have one master node each. We have plans on our roadmap to allow for multiple master nodes, keeping the Kubernetes API accessible during an upgrade and increasing the resilience in case of a machine failure.

I'm not sure, but isn't that what HA Masters was about? Does it need to be updated? https://docs.giantswarm.io/basics/ha-masters/

src/content/reference/cluster-upgrades/_index.md
paurosello: I think the original sentence was fine. 

A single machine will go through this steps and it's important to mention that a machine will be drained before recreation.

paurosello:
> Also, is this still true?
> 
> > **Note**: Currently tenant clusters have one master node each. We have plans on our roadmap to allow for multiple master nodes, keeping the Kubernetes API accessible during an upgrade and increasing the resilience in case of a machine failure.
> 
> I'm not sure, but isn't that what HA Masters was about? Does it need to be updated? https://docs.giantswarm.io/basics/ha-masters/

We should remove this too or maybe change it to: "Consider using HA master setup to avoid API downtime"

src/content/reference/cluster-upgrades/_index.md
```diff
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
 
 In particular this means:
 
-- Nodes will be drained, then stopped, then recreated.
+- Nodes are drained and recreated at the same time.
```
cokiengchiara, line 106:
```suggestion
- Nodes are drained and recreated at the same time.
```

src/content/reference/cluster-upgrades/_index.md
```diff
@@ -103,7 +103,7 @@ Your Kubernetes workloads should continue to work as expected (given they meet a
 
 In particular this means:
 
-- Nodes will be drained, then stopped, then recreated.
+Nodes will be drained, then stopped, then recreated.
```
cokiengchiara, line 106:
```suggestion
- Nodes will be drained, then stopped, then recreated.
```

src/content/reference/cluster-upgrades/_index.md
cokiengchiara: oh ok! put it back to the original ✌️

src/content/reference/cluster-upgrades/_index.md
```diff
@@ -107,7 +107,7 @@ In particular this means:
 - As a consequence of draining, Pods running on a node will be rescheduled to other nodes.
 - Once the master node is taken down and recreated, the Kubernetes API will be unavailable for a short time.
 
-**Note**: Currently tenant clusters have one master node each. We have plans on our roadmap to allow for multiple master nodes, keeping the Kubernetes API accessible during an upgrade and increasing the resilience in case of a machine failure.
+**Note**: By default, tenant clusters have one master node each. Consider using High Availability masters setup to avoid API downtime during upgrades. (See: [High availability Kubernetes masters](https://docs.giantswarm.io/basics/ha-masters/))
```
cokiengchiara, line 110: @paurosello WDYT?

paurosello: looks better!

src/content/reference/cluster-upgrades/_index.md
paurosello: This only applies if they are using single master

src/content/reference/cluster-upgrades/_index.md
marians: Thanks for spotting this! I missed that in my last edit.

You can even direct readers to `#checklist-ha-masters` here on the same page, not switching context too much. No need for parantheses in my opinion.

src/content/reference/cluster-upgrades/_index.md
cokiengchiara: can you suggest something Pau? I'm not really familiar with this area in the product :)

src/content/reference/cluster-upgrades/_index.md
cokiengchiara: @marians can you make the change please? :)

marians, line 110:
```suggestion
**Note**: By default, tenant clusters have one master node each. [Consider using a High Availability masters setup](#checklist-ha-masters) to avoid API downtime during upgrades.
```

marians: @cokiengchiara Here you go

cokiengchiara: danke!

src/content/reference/cluster-upgrades/_index.md
cokiengchiara: sounds like i can leave it as is.

cokiengchiara:
@paurosello or @marians can you check the changes and approve if looks ok? :)

marians (approved):
LGTM

PR #609, opened 2020-10-26
Add delete information

paurosello:
Add information for Cluster deletion via k8s API https://github.com/giantswarm/giantswarm/issues/11710

### DIFF
--- a/src/content/guides/creating-clusters-via-crs-on-aws/index.md
+++ b/src/content/guides/creating-clusters-via-crs-on-aws/index.md
@@ -222,7 +222,6 @@ As a result of rendering the CRs ([sample](/reference/kubectl-gs/template-cluste
 The resources can then be created by applying the manifest files to the Control Plane, e.g. `kubectl create -f <cluster manifest file>.yaml`.
 Of course, that requires the user to be authorized towards Kubernetes Control Plane API.
 
-
 ## How to delete a cluster using Cluster API
 
 In order to delete a cluster succesfully all the resources belonging to this instance need to be marked for deletion, this includes:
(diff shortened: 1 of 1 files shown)


src/content/guides/creating-clusters-via-crs-on-aws/index.md
```diff
@@ -221,3 +221,20 @@ There are also specific reference pages for [cluster templating](/reference/kube
 As a result of rendering the CRs ([sample](/reference/kubectl-gs/template-cluster/#example)), a user will get YAML manifests containing valid CRs that can create a tenant cluster and its node pools.
 The resources can then be created by applying the manifest files to the Control Plane, e.g. `kubectl create -f <cluster manifest file>.yaml`.
 Of course, that requires the user to be authorized towards Kubernetes Control Plane API.
+
+## How to delete a cluster using Cluster API
+
+In order to delete a cluster succesfully all the resources belonging to this instance need to be marked for deletion, this includes:
+
+- `Cluster` (API version `cluster.x-k8s.io/v1alpha2`)
```
xh3b4sd, line 229: It should be good enough to delete the `Cluster` CR. `cluster-operator` should propagate the deletion to all other CRs. This is also the behaviour of our conformance tests.

xh3b4sd (commented):
About docs structure, the changed file is about creation. Not sure we want to put deletion in there.

src/content/guides/creating-clusters-via-crs-on-aws/index.md
paurosello: hmm true, will move to different place and document that they can also only remove the cluster

PR #618, opened 2020-11-06
Improve fine-tuning upgrades guide

marians:
As talked about, this PR tries to improve several aspects at once.


src/content/guides/fine-tuning-upgrade-disruption-on-aws/index.md
```diff
@@ -1,123 +1,136 @@
...
+When a worker node update is necessary, nodes are updated (terminated and replaced by new ones) in groups or batches. The maximum batch size configures how many nodes are updated at the same time.
 
-### MaxBatchSize examples
+The settings directly maps to the [`MaxBatchSize`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-maxbatchsize) property of the AWS CloudFormation update policy.
 
-In this example, we set value to static value `10` to roll maximum of 10 instances per single batch.
+The **default value** for this property is `0.3`, which means that each batch will contain at most 30% of the nodes.
 
-```yaml
+In order to override the default for either the entire cluster or a spdcific node pool, the annotation
```
calvix, line 33: `spdcific ` -> specific

src/content/guides/fine-tuning-upgrade-disruption-on-aws/index.md
```diff
@@ -1,123 +1,136 @@
...
+
+must be set
+
+- on the [`AWSCluster`](/reference/cp-k8s-api/awsclusters.infrastructure.giantswarm.io/) resource, to be applied as a default to **all node pools** of the cluster.
+- on the [`AWSMachineDeployment`](/reference/cp-k8s-api/awsmachinedeployments.infrastructure.giantswarm.io/) resource, to be effective for only **one node pool**. A value here will override any value specified on the `AWSCluster` level.
+
+You have two options to configure the maximum batch size:
+
+- **Absolute**: using an integer number larger than zero, you'll specify the maximum number of nodes in absolute terms.
+- **Relative**: using a decimal number between `0.0` and `1.0` you can define the group size as a percentage of the node pool size. As an example, the value `"0.25"` would mean that all worker nodes would be divided into four groups, each containing 25 percent of the worker nodes (roughly).
```
calvix, line 47: maybe it's worth mentioning that they cannot use `0` but they can use `1.0`  (as 100% makes sense  and 0% dont)

you cannot roll zero nodes so the minimum number will always be 1 node

src/content/guides/fine-tuning-upgrade-disruption-on-aws/index.md
```diff
@@ -1,123 +1,136 @@
...
+- `PT1M30S`: 1 minute and 30 seconds
 
-The default value is `PT15M` (15 minutes).
+The maximum pause time is one hour (`PT1H`).
 
-### PauseTime examples
+### Examples {#pause-time-examples}
 
-In this example, we set the value to `PT1M30S` to pause for 1 minute and 30 seconds between each batch.
+In the first example, we set the value to `PT1M30S` to pause for one minute and 30 seconds between each batch.
```
calvix, line 108: ` one minute and 30 seconds` this feels weird to me that once we use the word and then we use the number, but if you feel its right then go with it

calvix (approved):

src/content/guides/fine-tuning-upgrade-disruption-on-aws/index.md
marians: I guess you're right. I was following the guideline "In scientific and technical writing, the prevailing style is to write out numbers under ten." here. But in direct combination, it would make more sense to use 1 instead of "one".

PR #619, opened 2020-11-09
Use HUGO version 0.78.1

marians:


ubergesundheit (approved):
LGTM if this is the only place to update

marians:
> if this is the only place to update

I think so 🤷

PR #629, opened 2020-11-16
Move kubectl gs template cluster '--master-az' flag to the 'Common' zone

axbarsan:
As seen in this card https://github.com/orgs/giantswarm/projects/82#card-48705179

This PR moves the flag to the `Common` flag zone, and adds info about how it works for `azure`


whites11 (approved):

tuommaki (approved):
Seems good to me.

PR #633, opened 2020-11-18
Fix the problem that the selected category was displayed in the wrong color

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/14391


oponder (approved):
LGTM!

PR #644, opened 2020-11-19
Add playground apps to aggregated change log

marians:
For https://github.com/giantswarm/giantswarm/issues/14418

This PR adds the repositories for our playground apps to the configuration (https://github.com/giantswarm/docs/commit/becbb2aded7c99515214abed4fc36b372cff6ebe) for the aggregated change log.

The additional file changes are mostly the addition of the releases of these apps.

## Preview

image

As we can see in the releases from `inlets/inlets`, there are some improvements to be made regarding the rendering of Github commit URLs. I'll keep that for a separate PR.

image


oponder (approved):

marians:
FYI: The commit hashes are linked now.

image

marians:
@cokiengchiara and @puja108 What are your thoughts regarding the categorization? (`Managed Apps` vs. `Playground Apps`)?

marians:
Alternative for a category name: `Experimental Apps`

puja108:
We are actually thinking that "Playground" (and "Experimental" even more) is underselling ourselves and at least with Henning and Oli we agreed that the level of support we factually provide for these apps is at least to be called "Supported". For Timo it is also more about "We have Apps" and then "some of them we offer full management with SLA for", there should not be big differences between the two besides the level of support we offer.

We have been working on a bit of clarification in following doc: https://docs.google.com/document/d/1IujvU9N2wESvQGnJ3wJcO5fYAMKSkdwaDWunefOzQ_M/edit

marians:
So another solution would be to put all apps into the same category and name it e. g. `Apps`.

cokiengchiara (approved):
LGTM if it works :D Thanks!

cokiengchiara:
> @cokiengchiara and @puja108 What are your thoughts regarding the categorization? (`Managed Apps` vs. `Playground Apps`)?

@puja108 is right and it's going in that direction. However, the naming and definitions is a separate decision. I don't think we should block this PR.

PR #648, opened 2020-11-19
Remove date from pages front matter, normalize last_review_date

marians:
Closes https://github.com/giantswarm/giantswarm/issues/14401

Changes the templates to use the last commit date from the git repo for each file.


marians:
This currently breaks in our build process.

image

I think that with the way we copy files from the original path to a different build path, we are missing the `.git` dir.

marians:
Closing, will start over

PR #650, opened 2020-11-20
Improvements to the Changes and Releases RSS feeds

marians:
- Adjusting the feed titles
- Make the full content of each release appear in the feed items, including markup

## Preview

From NetNewsWire

### Before

image

### After

image

image


src/layouts/changes/rss.xml
```diff
@@ -0,0 +1,39 @@
...
+    {{- with .OutputFormats.Get "RSS" }}
+    {{ printf "<atom:link href=%q rel=\"self\" type=%q />" .Permalink .MediaType | safeHTML }}
+    {{- end -}}
+    {{ range $pages }}
+    <item>
+      <title>{{ .Title }}</title>
+      <link>{{ .Permalink }}</link>
+      <pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
+      <guid>{{ .Permalink }}</guid>
+      <description>{{ .Content | html }}</description>
```
ubergesundheit, line 35: Why here `html` instead of `safeHTML`? Does it escape too much stuff?

marians:
Ehm, that wasn't supposed to happen. That automation should normally

- only work on the branch named `automated-update-for-changes`
- only happen on approval, not on a comment

I'll have to test this stuff in an extra repo...

src/layouts/changes/rss.xml
marians: html is safer.

Since we are scraping third party repos, we must not trust the input. safeHTML can only be used on trusted strings.

https://gohugo.io/functions/safehtml/

ubergesundheit: nice :+1:

PR #665, opened 2020-11-24
Fix code samples not rendering in Inconsolata.

oponder:
Fixes: https://github.com/giantswarm/giantswarm/issues/14489

This was due to a greedy style that was part of the automatically extracted header styles.


oponder:
@marians I'm a bit surprised about the colour difference between me and you though

Here's what it looks like in this branch right now:

Screenshot 2020-11-24 at 6 41 35 PM

axbarsan (approved):

marians:
> I'm a bit surprised about the colour difference between me and you though

Yes, that's indeed quite strange

PR #666, opened 2020-11-24
Fix lots of whitespace below the footer.

oponder:
This fixes a bunch of whitespace that appeared below the footer. It was coming from the base.css file we stored in the web assets repo. I don't think we need that in here at all anymore.

I think it was there originally to bring in the header and footer styles from the very old giant swarm website. When we changed the website, we removed the headers and footers, but kept referencing this file. I removed it, and I don't see anything breaking. 

------------------------------

before:

Screenshot 2020-11-24 at 10 33 20 PM

after:

Screenshot 2020-11-24 at 10 38 06 PM

### DIFF
--- a/src/assets/styles/base.sass
+++ b/src/assets/styles/base.sass
@@ -6,7 +6,6 @@
 body
   margin-top: 120px
 
-
 h1, h2, h3, h4, h5, h6
   text-align: left
   line-height: 1.3em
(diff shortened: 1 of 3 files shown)


src/assets/styles/base.sass
```diff
@@ -7,6 +6,7 @@
 body
   margin-top: 120px
 
+
```
axbarsan, line 9:
```suggestion
```

axbarsan (approved):

marians:
Strange I didn't see this... I don't understand software.

Anyway, great to shave off one HTTP request again :)

PR #672, opened 2020-11-27
Remove script intended for the header.

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/14578

I'll remove these scripts for now and re-introduce them in the PR that brings in the header.


marians (approved):
LGTM, but didn't test

PR #677, opened 2020-12-01
Add main site header

oponder:
This adds the header from the main site, unifying the look and feel of our docs and main website.

I also added a make target for copying the header and footer to their files in the partial folder, but we need one more for copying the mobile menu over too. Going into hubspot now to add the comments that `sed` can use to find the mobile header code.


src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+
+var headerLinks = document.getElementsByClassName("hs-menu-depth-1");
+for (var i = 0; i < headerLinks.length; i++) {
+  headerLinks[i].addEventListener('click', toggleMenu, false);
+}
+
+function toggleMenu(e) {
+  e.preventDefault();
+
+  e.currentTarget.classList.toggle("menu-expanded")
```
axbarsan, line 12:
```suggestion
  e.currentTarget.classList.toggle("menu-expanded");
```

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+
+  var wrapperContainer = e.currentTarget.getElementsByClassName('hs-menu-children-wrapper-container')[0]
```
axbarsan, line 14:
```suggestion
  var wrapperContainer = e.currentTarget.querySelector('.hs-menu-children-wrapper-container');
```

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+
+  if (wrapperContainer.style.display === "none" || ! wrapperContainer.style.display) {
```
axbarsan, line 16:
```suggestion
  if (wrapperContainer.style.display === "none" || !wrapperContainer.style.display) {
```

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+    closeAllMenus();
+    wrapperContainer.style.display = "block";
+  } else {
+    wrapperContainer.style.display = "none";
+  }
+}
+
+function closeAllMenus() {
+  for (var i = 0; i < headerLinks.length; i++) {
+    var wrapperContainer = headerLinks[i].getElementsByClassName('hs-menu-children-wrapper-container')[0];
```
axbarsan, line 26:
```suggestion
    var wrapperContainer = headerLinks[i].querySelector('.hs-menu-children-wrapper-container');
```

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+var mobileMenuTrigger = document.querySelector('.menu-trigger__container');
+var mobileMenu = document.querySelector('.mobile-menu');
+mobileMenuTrigger.addEventListener('click', function() {
+  mobileMenuTrigger.classList.toggle('open');
+  mobileMenu.classList.toggle('open');
+});
+
+function clickHandler(link) {
+  var l = link;
+  return (e) => {
```
axbarsan, line 44:
```suggestion
  return function(e) {
```

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+// Recreate open/close behaviour of individual menu items in the mobile menu.
+var mobileMenuLinks = document.querySelectorAll('.mobile-menu .hs-menu-depth-1 > a');
+for (var i = 0; i < mobileMenuLinks.length; i++) {
+  var link = mobileMenuLinks[i];
+
+  link.addEventListener('click', clickHandler(link));
+}
```
axbarsan, line 68:
```suggestion
}

```

src/layouts/partials/gs_styles.html
```diff
@@ -160,4 +169,9 @@
   .site-footer {
     margin-top: 100px;
   }
+
+  /* Fix Chevron Icon in the Mobile Menu */
+  .hs-menu-item .fa-chevron-down:before {
+    content: "\f078";
+  }
 </style>
```
axbarsan, line 177:
```suggestion
</style>

```

axbarsan (commented):
Looks good in general, just a few things about code style and a couple questions.

What do you think about running the menu code only when the `DOMContentLoaded` events fires, so all elements would be there?
Also, are we 100% sure that all the elements will always be present? We currently don't do any `null` checks on `querySelector()` results.

oponder:
> What do you think about running the menu code only when the DOMContentLoaded events fires, so all elements would be there?

> Also, are we 100% sure that all the elements will always be present? We currently don't do any null checks on querySelector() results.

I think we can be sure they will be there and don't need to rely on the DOMContentLoaded event. Since the script is loaded at the end, before the body tag closes. All the elements should be there by then. 

Also header, footer, and the mobile menu should be present on every page.. otherwise we have a problem. 

But if you think it's a better code style / better practice thing I'm happy to do that.

Makefile
```diff
@@ -109,6 +109,10 @@ linkcheck-external:
 	docker kill server
 	docker kill linkchecker
 
+grab-main-site-header-footer:
+	curl -s https://www.giantswarm.io/why-giant-swarm | sed -n '/^<!-- BEGIN SITE_HEADER -->/,/^<!-- END SITE_HEADER -->/p;/^<!-- END SITE_HEADER -->/q' | sed '1d;$d' > src/layouts/partials/gs_header.html
+	curl -s https://www.giantswarm.io/why-giant-swarm | sed -n '/^<!-- BEGIN SITE_FOOTER -->/,/^<!-- END SITE_FOOTER -->/p;/^<!-- END SITE_FOOTER -->/q' | sed '1d;$d' > src/layouts/partials/gs_footer.html
```
ubergesundheit, line 114: Can you add a section about to use this to the `README` or `CONTRIBUTING` file?

Also can we add a little safety here? I'm thinking about cases where curl fails, or returns some unexpected output.

Maybe just a check that the file size is not below a certain value? For example with `[ "$(wc -c < src/layouts/partials/gs_header.html)" -ge 2000 ]` (2000 is arbitrary)
Maybe just a check

axbarsan (approved):

Makefile
oponder: I don't wanna over think this part right now, also don't understand the check yet. I think maybe we can take this in another PR that tries to automate this further. For now it's like a little hint to the future me or whoever takes it on next, as to what gs_header and gs_footer really contain.

oponder:
ping @marians 

What do you think about the usability of this header in the docs context?

Once you scroll down, you lose the docs nav, and you are forced to think "how did I get here, and how do I go back to where I was".

You have to scroll all the way up to see the main docs navigation. This isn't any different from how it is already before this PR.

However now with more things to consider in the header, you might get lost for a while trying to find the way back to the docs root in the main giant swarm nav.

Screenshot 2020-12-02 at 10 50 47 PM

marians:
I'm not too concerned about the main header being sticky while the docs menu is disappearing. We had this for quite some time and it didn't seem to bother too much. Anyway, something to iterate on, maybe? Then it should perhaps be a site-wide thing, not just docs.

marians (changes requested):
A functional blocker: links in the menu don't work for me. I click on `Company`, then on any other link, e. g. `Careers`, and the menu closes again instead of taking me to the target URL.

oponder:
Nice! Thanks for testing that. I guess I just assumed it worked 😅

src/assets/scripts/gs_menu.js
```diff
@@ -0,0 +1,69 @@
+// This recreates the behaviour of the Giant Swarm menu without copying over all
+// of that site's javascript.
+
+var headerLinks = document.querySelectorAll(".hs-menu-depth-1 > a");
+for (var i = 0; i < headerLinks.length; i++) {
+  console.log(headerLinks[i]);
+  headerLinks[i].addEventListener('click', toggleMenu, false);
```
marians, line 7:
```suggestion
for (var i = 0; i < headerLinks.length; i++) {
  headerLinks[i].addEventListener('click', toggleMenu, false);
```

marians:
How did you extract the code now? I looked into the menu content again and I'm expecting it to change frequently. I'm thinking that this only makes sense if we are able to update the content quickly and basically completely automatically.

marians (approved):

marians:
Ping @oponder - The outcome works well for me. I had only one question about the updating workflow. See above.

oponder:
@marians I've added some little curl commands to the Makefile which can serve as a starting point for the next person that continues working on this (could be me, but just wanted to be able to park it for now)

As for the CSS extraction, that's not anywhere yet, It's just not packaged in any useful form as it is right now. But I'm happy to share that if you are itching to work on the automation step. Otherwise I will probably try to continue from there when time allows.

marians:
OK, let's move one then :)

PR #679, opened 2020-12-02
Added questions to metadata

Oshratn:
Towards #13270
Also changed to American spelling, as is the standard at GS.
Question: How does the difference in American and British spelling affect search?


ubergesundheit (approved):
LGTM. No Idea on search

marians:
> Question: How does the difference in American and British spelling affect search?

It does affect search results in fact. For example, try https://docs.giantswarm.io/search/?q=organisation vs. https://docs.giantswarm.io/search/?q=organization

Oshratn:
So.. should in cases with UK and US spelling, should there be 2 questions in the front matter?

marians:
> So...in cases with UK and US spelling, should there be 2 questions in the front matter?

No, I think that would be the wrong way to handle this. Let's dig into technical ways to solve this, at some point. I'll create a new issue.

marians:
@Oshratn I think this is ready to merge

PR #683, opened 2020-12-03
Use last modified date from git, remove front matter date

marians:
Second attempt for https://github.com/giantswarm/docs/pull/648

This removes the front matter field `date` from all pages that are created/maintained manually. No more need to keep this up-to-date manually.


Dockerfile
```diff
@@ -4,9 +4,16 @@ RUN apk --no-cache add findutils gzip
 
 WORKDIR /docs
 
-COPY src .
-
-RUN hugo --verbose --gc --minify --cleanDestinationDir --path-warnings --destination /public
+COPY . /docs
```
marians, line 7: Copies the entire repo tree, including the `.git` dir.

Dockerfile
```diff
@@ -4,9 +4,16 @@ RUN apk --no-cache add findutils gzip
...
+
+RUN hugo \
+      --verbose \
+      --gc \
+      --minify \
+      --source src \
```
marians, line 13: Added `--source src` and some line breaks

src/config.toml
```diff
@@ -3,6 +3,7 @@ languageCode = "en"
 title = "Giant Swarm Documentation"
 pluralizeListTitles = false
 rssLimit = 50
+enableGitInfo = true
```
marians, line 6: This enables the feature

axbarsan (approved):
Niceee 👍🏻 please also announce this on Slack after you merge it

PR #685, opened 2020-12-03
Improve search result rankings

marians:
Add ranking function that reduces the score for changelog items and Rest API reference items. This is to improve the experience for queries where many changelog items can appear in the search result. E. g. [`release`](https://docs.giantswarm.io/search/?q=release).

Closes https://github.com/giantswarm/giantswarm/issues/14577


oponder (approved):
LGTM!

PR #694, opened 2020-12-04
Assign docs article ownership to Team Firecracker

marians:
Towards https://github.com/giantswarm/giantswarm/issues/13382

### DIFF
--- a/src/content/basics/nodepools/index.md
+++ b/src/content/basics/nodepools/index.md
@@ -9,6 +9,8 @@ user_questions:
   - What are node pools?
   - In which cloud environments are node pools supported?
   - Which tenant cluster releases introduced node pools?
+owner:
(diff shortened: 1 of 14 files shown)


src/content/basics/nodepools/index.md
```diff
@@ -9,6 +9,8 @@ user_questions:
   - What are node pools?
   - In which cloud environments are node pools supported?
```
anvddriesch, line 10: I'm not blocking this, just wondering whether this should be just firecracker, since we also support nodepools in azure?

anvddriesch (approved):

src/content/basics/nodepools/index.md
marians: We discussed this in SIG product and Alex and Marta decided that Firecracker will own it anyway.

PR #724, opened 2020-12-11
Correct OIDC yaml example

pipo02mix:


xh3b4sd (approved):
I am not sure where the annotation approach was coming from earlier. Though if this is how it works on AWS, which I think it is, then this should be good. 😄

PR #726, opened 2020-12-11
Add metadata for Spark CRD

marians:
This PR adds metadata for the Spark CRD, so that the tags `tenantcluster` and `AZURE` appear.

@tuommaki We also have a `controlplane` tag, and we could add both of both would make sense. I can't judge this, so your call.

### DIFF
--- a/src/data/crd_metadata.yaml
+++ b/src/data/crd_metadata.yaml
@@ -136,6 +136,11 @@ releases.release.giantswarm.io:
 silences.monitoring.giantswarm.io:
   topics:
     - controlplane
+sparks.core.giantswarm.io:
+  provider:
+    - azure
+  topics:
+    - tenantcluster
 storageconfigs.core.giantswarm.io:
   topics:
     - controlplane


tuommaki:
What meaning these tags has? How do they work or in other words, what implications do they have?

marians:
They don't have a 100% clearly defined meaning. The only function they have is to appear in the listing on https://docs.giantswarm.io/reference/cp-k8s-api/ 

They should help users understand "about" what a CRD is. `app` is about managed apps. `controlplane` is about the control plane/installation. `tenantcluster` influences tenant clusters.

tuommaki (approved):
I think tags are mostly up to you.

Some additional info that might help:

This CR object is created on `controlplane` but it's purpose is `tenantcluster` and for each TC machine type there is one `Spark` instance. In Azure the TC master node has one spark object and each node pool has own spark object. It's effectively 1:1 mapping between different kinds of ignition templates needed for tenant cluster.

Given that `silences` above has been tagged for `controlplane`, then I _guess_ this shouldn't have that tag because this doesn't modify CP in any way.

marians:
Thanks for the additional info! I guess the tag purpose needs to be defined in more details. Anyway, not a lot of harm will be done with one or two questionable tags here.

PR #738, opened 2021-01-05
Add info on skipping releases in upgrades

marians:
Adding instructions on how to skip a tenant cluster release in our UIs


oponder (approved):
LGTM

PR #742, opened 2021-01-06
Added search questions and boyscouting

Oshratn:
I added search questions towards issue #13270 in GS/GS.
I removed references to control plane, except where Control Plane Kubernetes API. As naming is pending in the workgroup.
I removed the reference to the Helm stable catalog, since IIRC this option was removed.


src/content/basics/app-platform/index.md
```diff
@@ -9,13 +9,23 @@ aliases:
   - /basics/app-catalog/
 owner:
   - https://github.com/orgs/giantswarm/teams/team-batman
+user_questions:
+  - What does App Platform do?
+  - What is an App Catalog?
+  - Does Giant Swarm provide app catalogs out of the box?
+  - How is App Platofrm implemented?
```
rossf7, line 16:
```suggestion
  - How is App Platform implemented?
```

src/content/basics/app-platform/index.md
```diff
@@ -9,13 +9,23 @@ aliases:
...
+  - What are the components of App Platform?
+  - What are the app catalogs that Giant Swarm provides out of the box?
+  - How can I create an organizational app catalog?
+  - How can I interact with the Giant Swarm App Platform?
+  - Can I create an app catalog?
```
rossf7, line 21: What is the different between this and "How can I create an organizational app catalog?" ?

rossf7 (approved):
Nice, just one question.

Also I think good to remove Helm stable. It's deprecated and we're removing after 6 months IIRC. But its not heavily used and we don't want to encourage its usage.

src/content/basics/app-platform/index.md
Oshratn: Good point. I was a bit torn between adding both. 
What ultimately won the day is that one is about the possibility and the other is how to undertake the task. Since these are search-related questions, I'm not completely sure a reader searching would default to how to undertake the task, since s/he might not be clear that it is even an option. If that makes sense.

rossf7: > What ultimately won the day is that one is about the possibility and the other is how to undertake the task.

OK cool I think I get this. So organizational app catalog Q is because we want to make this possible.

But I'm still seeing 3 references to control plane. I think these should also change to management cluster. Can you check again?

Oshratn: Yes, those are still the references to Control Plane API. As I mentioned I am waiting on word from the WG to see what they ultimately decide to rename it to.

rossf7: Ah OK I missed that. Thanks for clarifying. 👍

marians (approved):

PR #748, opened 2021-01-12
Add 'kgs get nodepools' docs

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14854


src/content/reference/kubectl-gs/get-clusters.md
```diff
@@ -79,7 +79,7 @@ kgs get clusters ab12c --output yaml
 When applied without a cluster ID argument, the output will be a list of resources. Example:
 
 ```nohighlight
-$ kgs get clusters ab12c --output yaml
```
axbarsan, line 82: This was incorrect

oponder: Thx for noticing and fixing!

src/content/reference/kubectl-gs/get-nodepools.md
```diff
@@ -0,0 +1,94 @@
+---
+title: "'kubectl gs get nodepools' command reference"
+description: Reference documentation on how to list node pools and get details for a single node pools using 'kubectl gs'.
```
oponder, line 3:
```suggestion
description: Reference documentation on how to list node pools and get details for a single node pool using 'kubectl gs'.
```

oponder: An accidental plural "node pools" in a singular context

pipo02mix (approved):
LTGM

src/content/reference/kubectl-gs/get-nodepools.md
```diff
@@ -0,0 +1,94 @@
...
+- `NODES READY`: The number of nodes in the node pool that are actually ready.
+- `DESCRIPTION`: User friendly description for the node pool.
+
+## Flags {#flags}
+
+Here we document the flags that have a particular meaning for the `get nodepools` command. Use `kgs get nodepools --help` for a full list.
+
+### `--cluster-id/-c` {#flags-output}
+
+If present, list the node pools that belong to this given workload cluster.
```
oponder, line 64: This triggered some thinking about the word `given` which I shared in sig-docs.  I don't think it has to be changed here though

src/content/reference/kubectl-gs/get-nodepools.md
```diff
@@ -0,0 +1,94 @@
...
+
+### `--output/-o` {#flags-output}
+
+`kubectl` commonly allows to specify the output format for all `get` subcommands. `kgs get nodepools` is no different.
```
oponder, line 68: > `kubectl` commonly allows to specify

I feel like we often use this grammatical sentence structure in our writing, and I think it's actually not right. I feel like the following would be more grammatically correct:

> kubectl allows _you_ to specify

or

> kubectl allows for specification of ...

I don't know if allow is even the best word in most cases where we use it. 
"Allow" implies permission and security or authorisation. 

In most cases where we use 'allow' we want to say that this is a capability or feature of the tool.

So I think "can" might be a better word.

> kubectl can specify the output format for...

oponder: Also this sentence is really about how this command is no different from other `get` commands when it comes to output formatting. It's setting up a comparison between two things and saying that they are the same or similar. 

So perhaps we can  arrive quickly at the core idea of this sentence like:

"Similar to other `get` subcommands, you can specify the output format of `kgs get nodepools` using the `-o` flag"

oponder (approved):
LGTM, left some notes because I got thinking about some things. But feel free to decide on the scope of the changes you want to make yourself.

src/content/reference/kubectl-gs/get-nodepools.md
```diff
@@ -0,0 +1,94 @@
...
+ID      CLUSTER ID   CREATED                         CONDITION   NODES MIN/MAX   NODES DESIRED   NODES READY   DESCRIPTION
+ab12c   s921a        2021-01-02 15:04:32 +0000 UTC   READY       3/10            5               3             Production node pool
+```
+
+### Get specific node pool
+
+When used with a node pool ID as additional argument, the command will show details for a single node pool. Example:
+
+```nohighlight
+kgs get nodepools ab12c
```
marians, line 40: I would prefer singular here

```suggestion
kgs get nodepool ab12c
```

src/content/reference/kubectl-gs/get-nodepools.md
```diff
@@ -0,0 +1,94 @@
...
+Note: As an alternative to `get nodepools`, `get nodepool` will also work.
+
+## Output {#columns}
+
+The standard tabular output format features these columns:
+
+- `ID`: Unique identifier of the node pool.
+- `CLUSTER ID`: Unique identifier of the cluster that the node pool belongs to.
+- `CREATED`: Date and time of the node pool CR creation.
+- `CONDITION`: Latest condition reported for the node pool.
```
marians, line 52: As explained in https://github.com/giantswarm/kubectl-gs/pull/254#issuecomment-760955591, this is not available on AWS. That could be mentioned here. Also we could explain what `n/a` could mean on Azure.

PR #762, opened 2021-01-25
Ignore deprecated CRDs

MartaD:
Adding CRDs to this list will omit them when docs are generated


stone-z (approved):
LGTM but just to double check -- are the names singular like this? It might be `drainerconfigs`?

MarcelMue:
I think it should be:
```
drainerconfigs.core.giantswarm.io
storageconfigs.core.giantswarm.io
```

MartaD:
How can I test it? Do I need to test it?

stone-z (approved):
That looks better to me

whites11 (approved):

PR #763, opened 2021-01-25
Make list styles more specific

marians:
This is a cheap attempt to mitigate https://github.com/giantswarm/giantswarm/issues/14826

However it's not the right approach in terms of maintainability. Because it means that these tweaks have to be repeated the next time we want to update the header and footer styles.

Can we somehow ensure that all styles imported for the header and footer are prefixed by `.site-header, site-footer` so that they only apply to these containers? The side effects can otherwise be unpredictable.

Side notes: Also I noticed there is a lot of redundancy in the CSS we import. And I don't like the fact that we include it as inline CSS in every page instead of having it be part of a CSS file we load once.

So please see this PR as a discussion starter. @ubergesundheit @oponder


marians:
Could @giantswarm/chapter-front-end please take a look here?

ubergesundheit (approved):
If the changes work for you :+1:

PR #768, opened 2021-01-28
Mention v12.0.0 regarding preinstalled nginx ingress controller

ubergesundheit:
Towards giantswarm/giantswarm#15428

Ping @giantswarm/team-celestial is this correct?


whites11 (approved):
correct for azure

PR #770, opened 2021-01-28
Fix info on default max number of nodes

marians:
The page currently shows this:

> When creating a cluster without specifying the number of worker nodes, 3 worker nodes will be created. On supported AWS and Azure workload clusters, when not specified, the maximum number of worker nodes is also set to 3.

The current default max is 10, not 3.


whites11 (approved):
thanks

calvix (approved):

PR #781, opened 2021-01-31
Update index.md

Oshratn:
Updated question for help. Towards issue: https://github.com/giantswarm/giantswarm/issues/13270

### DIFF
--- a/src/content/basics/multiaz/index.md
+++ b/src/content/basics/multiaz/index.md
@@ -4,6 +4,7 @@ description: Using multiple availability zones both for worker and for master no
 weight: 100
 type: page
 categories: ["basics"]
+last_review_date: 2020-01-02
 user_questions:
 - Does Giant Swarm support multiple availibilty zones (AZ)?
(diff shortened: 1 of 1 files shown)


ubergesundheit (approved):
LGTM

whites11 (approved):
Yeah makes sense

src/content/basics/multiaz/index.md
```diff
@@ -4,7 +4,17 @@ description: Using multiple availability zones both for worker and for master no
 weight: 100
 type: page
 categories: ["basics"]
-last_review_date: 2020-04-09
```
marians, line 7: Why do you delete the last_review_date?

Oshratn: They seem to have disappeared from other docs and I'm not sure how usable this field will be for us.
Happy to hear thoughts about the necessity of the field.

marians: We only had it on very few documents, since we haven't introduced the review process yet. I would leave it where we have it.

PR #786, opened 2021-02-02
Bump crd-docs-generator to 0.3.0 & add updated reference pages

ubergesundheit:
Towards giantswarm/giantswarm#15326

This PR bumps the version of crd-docs-generator to 0.3.0 which contains changes to the used template for rendering yaml code blocks.

It uses tripple backticks instead of `<pre><code>` markup. (Related giantswarm/crd-docs-generator#20)

It also deleted `storageconfigs.core.giantswarm.io`, `drainerconfigs.core.giantswarm.io` and `etcdbackups.backup.giantswarm.io` reference pages. Is this correct?

<details>

<summary>Screenshot</summary>

image

</details>


src/content/reference/management-api/drainerconfigs.core.giantswarm.io.md
```diff
@@ -1,357 +0,0 @@
----
```
ubergesundheit, line 1: Is this deletion (and others) ok?

stone-z: Yes, these were intentionally excluded (drainerconfigs and storageconfigs are deprecated, etcdbackups are internal)

src/content/reference/management-api/appcatalogs.application.giantswarm.io.md
```diff
@@ -37,7 +37,8 @@ aliases:
 
 
 <h3 id="crd-example-v1alpha1">Example CR</h3>
-<pre class="crd-example-cr"><code class="language-yaml">
```
stone-z, line 40: Is this class needed for anything else?

stone-z (approved):
This is nicer to work with anyway. If it renders fine then :rocket:

src/content/reference/management-api/drainerconfigs.core.giantswarm.io.md
stone-z: For reference: https://github.com/giantswarm/giantswarm/issues/15025

src/content/reference/management-api/appcatalogs.application.giantswarm.io.md
ubergesundheit: Looks like it was only in the html, no styling done with these classes (`crd-example-cr` and `language-yaml`)

src/content/reference/management-api/drainerconfigs.core.giantswarm.io.md
ubergesundheit: Although to not break external links, we should consider keeping the pages around, clearly marking them as deprecated but not link them from the list.

marians:
Thank you!

PR #789, opened 2021-02-03
Include External DNS in Changes and Releases

cokiengchiara:
Towards https://github.com/giantswarm/roadmap/issues/132

In addition to being a managed preinstalled app, external dns can now be used as a managed optional app.


glitchcrab (approved):
:partying_face:

PR #810, opened 2021-02-11
Add user questions to docs pages owned by SIG UX

marians:


axbarsan (approved):
What are these for?

marians:
Sorry for the lack of context! The epic is https://github.com/giantswarm/giantswarm/issues/13270

We plan to have a special sort of search function in docs that helps users find _the_article answering their questions. That's what we need those for.

PR #850, opened 2021-02-24
Add cloud provider tagging in Cluster custom resource

pipo02mix:
### Please remember to

- [ ] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/master/CONTRIBUTING.md#front-matter) associated with any updated docs

### DIFF
--- a/src/content/ui-api/management-api/creating-workload-clusters/aws/index.md
+++ b/src/content/ui-api/management-api/creating-workload-clusters/aws/index.md
@@ -256,7 +256,7 @@ __Note__: Currently changing/adding config to already existing cluster is not fu
 
 ## Cluster provider resource tagging
 
-Since AWS Giant Swarm release `14.0.0` our operator will propage cluster CR labels to the cloud provider resources (if they have the prefix `tag.provider.giantswarm.io/`). Take into account that it will only work, for most of the (static) resources, when you set the labels at creation time. 
+Since AWS Giant Swarm release `14.0.0` our operator will propage cluster CR labels to the cloud provider resources (if they have the prefix `tag.provider.giantswarm.io/`). Take into account that it will only work, for most of the (static) resources, when you set the labels at creation time.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


anvddriesch (approved):

src/content/ui-api/management-api/creating-workload-clusters/aws/index.md
```diff
@@ -44,6 +44,7 @@ kind: Cluster
 metadata:
   generation: 1
   labels:
+    tag.provider.giantswarm.io/Environment: PROD
```
njuettner, line 47: one question regarding labels, did we ever had an agreement keeping labels in lower case?

njuettner: the reason why I'm asking because kubernetes labels are case-sensitive

njuettner: also tag.provider.giantswarm.io/Environment ->  tag.provider.giantswarm.io/environment

pipo02mix: well this is up to the customer. We just define the prefix `tag.provider.giantswarm.io` and they will decide if they want to use lower or uppercase. We could define some guidelines but not enforce them, right?

njuettner: fully agree, we should have some guidelines around labels. It has some implications when using label selector and stuff so customer has to be careful in case it uses those labels to select. Not blocking here and thanks for updating the docs 🚀 .

njuettner (approved):

src/content/ui-api/management-api/creating-workload-clusters/aws/index.md
marians: I would also prefer giving an all lowercase example. Seems to me as a good habit to not involve uppercase/lowercase selection issues with labels.

PR #854, opened 2021-02-26
Update "Establishing trust to your cluster's CA and importing certificates" tutorial for Firefox on macOS

marians:
This change simplifies the tutorial for Firefox on macOS based on the fact that Firefox supports the system keychain since version 75.

See

- https://blog.mozilla.org/security/2020/04/14/expanding-client-certificates-in-firefox-75/
- https://support.mozilla.org/en-US/kb/setting-certificate-authorities-firefox

In addition, the spelling of `macOS` is aligned.

### Please remember to

- [ ] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/master/CONTRIBUTING.md#front-matter) associated with any updated docs


marians:
@giantswarm/team-ludacris Pinging you for review here as you are owning this article.

According to the sources mentioned above, the same method that is explained here for Firefox on macOS should also be supported on Windows and it may work on some Linux versions, too. However I cannot verify this easily myself now. Also that shouldn't block the change IMO.

stone-z (approved):
I'm not able to personally test the macOS functionality, but looks reasonable to me

PR #863, opened 2021-03-09
Adding App metadata validation

tomahawk28:
### Please remember to


src/content/app-platform/defaulting-validation/index.md
```diff
@@ -109,6 +109,11 @@ Currently we validate:
 - All referenced configmaps and secrets exist.
 - The catalog has a matching AppCatalog CR.
 
+If app-operator finds a matching AppCatalogEntry CR named `{appCatalog}-{app}-{version}`, it will run more two validation.
```
rossf7, line 112:
```suggestion
If app-operator finds a matching [AppCatalogEntry]({{< relref "/ui-api/management-api/crd/appcatalogentries.application.giantswarm.io.md" >}}) CR, it will use this to run more validation checks.
```

I don't think we need to show the naming convention but we could link to the CRD docs. I just removed the two as we may add more checks in the future.

src/content/app-platform/defaulting-validation/index.md
```diff
@@ -109,6 +109,11 @@ Currently we validate:
...
+
+- Cloud provider compatibility (e.g. you can’t install kiam app in azure, for example)
```
rossf7, line 114:
```suggestion
- Cloud provider compatibility (e.g. you can’t install the azure-ad-pod-identity app in AWS).
```

kiam is pretty well know but I just switched the example because it has Azure in the name. WDYT?

src/content/app-platform/defaulting-validation/index.md
```diff
@@ -109,6 +109,11 @@ Currently we validate:
...
+- Namespace restriction (cluster singleton, namespace singleton, fixed namespace) (e.g You can only install cert-manager in the kube-system namespace.)
```
rossf7, line 115:
```suggestion
- Namespace restriction (cluster singleton, namespace singleton, fixed namespace) (e.g You can only install cert-manager in the kube-system namespace).
```

I think the full stop should be at the end. I think the usage of singleton so we don't need to explain that further.

src/content/app-platform/defaulting-validation/index.md
tomahawk28: Nice that we put the link on the CRD document. Thanks!

src/content/app-platform/defaulting-validation/index.md
tomahawk28: Good catch, let's use azure app instead.

src/content/app-platform/defaulting-validation/index.md
tomahawk28: >  I think the usage of singleton so we don't need to explain that further. 

Sorry, I don't get this part, do you think we should remove the explanation part of this?

rossf7: My bad I meant to say.

I think the meaning of singleton is clear so we don't need to explain that further.

rossf7 (approved):
LGTM once those changes are in. 👍

PR #904, opened 2021-04-19
Add blog posts to search UI

marians:
This PR adds searching in the blog index, potentially returning hits for blog posts matching the search query.

Towards https://github.com/giantswarm/giantswarm/issues/15387

## Preview

image

### DIFF
--- a/src/assets/scripts/base.js
+++ b/src/assets/scripts/base.js
@@ -185,13 +185,13 @@ function doSearch(q) {
  */
 function renderSerpEntry(index, hit) {
   d = $("<a class='item'></>").attr("href", hit._source.uri);
-  if (typeof(hit.highlight) !== "undefined" && typeof(hit.highlight.title) !== "undefined") {
+  if (typeof hit.highlight !== "undefined" && typeof hit.highlight.title !== "undefined") {
     d.append($("<h4></h4>").html((index + 1) + ". " + hit.highlight.title));
   } else {
     d.append($("<h4></h4>").html((index + 1) + ". " + hit._source.title));
   }
 
-  if (typeof(hit._source.breadcrumb) !== "undefined" && hit._source.breadcrumb.length > 0) {
+  if (typeof hit._source.breadcrumb !== "undefined" && hit._source.breadcrumb.length > 0) {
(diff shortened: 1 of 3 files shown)


Oshratn:
LGTM - I like the fact that it is clear that the hit is a blog post and the publish date.

When there is a mixed bag - how will it appear? By date? By relevance? Broken up into sections (i.e. results from docs and results from blog)?

Oshratn (approved):

marians:
> When there is a mixed bag - how will it appear? By date? By relevance? Broken up into sections (i.e. results from docs and results from blog)?

Right now, it's mixed (so that the changes to the search UI are minimal). The sorting is by relevance. Relevance is a bit more complex. Blog articles are by default less relevant than normal docs articles. But if a blog post fits the search entry much better than a docs article, it will come up higher.

src/assets/scripts/base.js
```diff
@@ -171,9 +190,23 @@ function renderSerpEntry(index, hit) {
   } else {
     d.append($("<h4></h4>").html((index + 1) + ". " + hit._source.title));
   }
+
   if (typeof(hit._source.breadcrumb) !== "undefined" && hit._source.breadcrumb.length > 0) {
-    d.append($("<p class='sbreadcrumb'></p>").html("/" + hit._source.breadcrumb.join("/") + "/"));
+    var top = hit._source.breadcrumb[0].toLowerCase();
+    if (top === 'blog') {
+      var breadcrumb = 'Blog post'
+      if (typeof(hit._source.published) !== "undefined") {
```
ubergesundheit, line 198: Did this work? IMO the syntax to check for undefined is like this:

```suggestion
      if (typeof hit._source.published !== "undefined") {
```

ubergesundheit (approved):
LGTM.

Regarding the `typeof` syntax it seems I was wrong. Both with and without parentheses is supported. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

marians:
> Regarding the typeof syntax it seems I was wrong.

😱 

Thanks for letting me know :)

PR #926, opened 2021-05-03
Linkcheck update

marians:
- Make the `make linkcheck` target more usable by using the live website instead of a container, storing the report as a CSV file, using only one thread instead of 2.
- Fix a URL.


ubergesundheit (approved):
LGTM

PR #933, opened 2021-05-05
Update sample PSP and ClusterRoles to use policy API group

stone-z:
Fixes https://github.com/giantswarm/giantswarm/issues/17088

### Please remember to


marians:
Wow, that was quick!

Did you check whether that works with all active Kubernetes version in our releases in use?

stone-z:
PSPs were moved from `extensions` to `policy` in v1.16.0. We still have a few (5ish) clusters on v1.15.x, but the new API group is used by the vast majority. I'm not sure it's very useful to include a caveat for the 1.15 stragglers as I would suspect/hope most are in a sunset phase already. But I'm happy to add one if you think we should

Oshratn:
I think what @stone-z makes sense.
If we do have a 1.15 straggler maybe the omission would cause the reader to contact us, which will help us trigger an upgrade.

marians (approved):

PR #934, opened 2021-05-05
Clarify difference between App Platform and App Catalogs in web UI docs

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/16903

Follow up to https://github.com/giantswarm/docs/pull/917

@Oshratn asked if we can make clearer the difference between App Platform and App Catalogs. I also added a note on how Helm relates to App Platform. 

We should also update this page after the navigation changes. The list of catalogs is no longer shown. @oponder Could you take a look at that?

### Please remember to


Oshratn (approved):
LGTM

rossf7:
> But it seems that @rossf7 is looking into a more in-depth rewrite.

The main problem is this screenshot is outdated. We should update it to use the new search-based navigation.

app-catalogs

Going with this for now.

PR #940, opened 2021-05-11
Rank down older search result entries

marians:
Towards https://github.com/giantswarm/giantswarm/issues/16853

This re-enables the date based decay function (which had been added recently and then removed again in https://github.com/giantswarm/docs/pull/908) to decrease the score of result entries based on their age.

With https://github.com/giantswarm/docs-indexer/pull/40 all index entries will have a date field, so this actually works.

In addition, this improves the display of changelog/release entries in the search results.

## Preview

image


stone-z (approved):
LGTM

PR #986, opened 2021-06-22
Update docs for kubectl gs template app

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/16978

Code: https://github.com/giantswarm/kubectl-gs/pull/379

The current `kubectl gs template app` sets a lot of boilerplate fields that can be defaulted. It also has ugly empty fields due to a problem with the App CRD.

### Please remember to


src/content/ui-api/kubectl-gs/template-app.md
```diff
@@ -8,6 +8,7 @@ menu:
     parent: uiapi-kubectlgs
 aliases:
   - /reference/kubectl-gs/template-app/
+last_review_date: 2021-02-22
```
marians, line 11:
```suggestion
last_review_date: 2021-06-22
```

src/content/ui-api/kubectl-gs/template-app.md
```diff
@@ -47,41 +48,30 @@ produces the following output:
...
-  appVersion: ""
-  release:
-    lastDeployed: "0001-01-01T00:00:00Z"
-    status: ""
-  version: ""
 ```
+
+It also supports the following optional flags:
+
+- `--defaulting-enabled`: Don't template fields that will be defaulted. (default true)
```
marians, line 64: Instead of a negative phrase like "Don't template fields" I'd rather explain what this does in a positive way. E. g. "Only include details that differ from default value. When false, a much longer template is created."

src/content/ui-api/kubectl-gs/template-app.md
```diff
@@ -47,41 +48,30 @@ produces the following output:
...
+- `--user-configmap`: Path to the user values configmap YAML file.
+- `--user-secret`: Path to the user secrets YAML file.
+
+Only required fields are templated. Other fields like the path to the kubeconfig secret are set by the
+[defaulting webhook]({{< relref "/app-platform/defaulting-validation" >}}).
```
marians, line 69: If I understand this correctly, this behaviour is overwritten via `--defaulting-enabled=false`. If that's the case, I think we should mention it somehow.

src/content/ui-api/kubectl-gs/template-app.md
marians: Further down I read

> For older releases you can set the `--defaulting-enabled` flag to false.

Is this flag effective for all releases, or just for some?

src/content/ui-api/kubectl-gs/template-app.md
```diff
@@ -47,41 +48,30 @@ produces the following output:
...
+
+This is enabled for Giant Swarm releases.
+
+- AWS >= v14.0.0
+- Azure >= v13.1.0
+- KVM >= v13.1.0
+
+For older releases you can set the `--defaulting-enabled` flag to false.
```
marians, line 77: Ah, here it is. As mentioned, I think I'd like this detail a bit further up near `Only required fields are templated.`.

src/content/ui-api/kubectl-gs/template-app.md
rossf7: Yes if the webhook is not enabled you can still use template app by setting this to false. I moved it up higher as yes that flows better.

src/content/ui-api/kubectl-gs/template-app.md
rossf7: Thanks, that's much better.

marians (approved):

tomahawk28 (approved):

rossf7:
Changes released in kubectl-gs v1.30.0

https://github.com/giantswarm/kubectl-gs/releases/tag/v1.30.0

PR #988, opened 2021-06-23
docker-compose: remove restart policy from sitesearch

marians:
This is a tiny internal development change. The goal is to prevent sitesearch (Elasticsearch) from starting each time I start Docker engine locally.


ubergesundheit (approved):
yup!

PR #989, opened 2021-06-23
Fail in CI if front matter is not okay

marians:
Towards https://github.com/giantswarm/giantswarm/issues/15542

This PR adds a github workflow to validate the front matter on every push into a branch. Preview of a non-okay result:

image


marians:
@ubergesundheit After a little journey this is ready for review. There is stilly a `dummy.md` file in here for demo purposes which I will of course delete before merging.

.circleci/config.yml
```diff
@@ -16,10 +16,6 @@ jobs:
           name: Check Markdown structure
           command: make lint
 
-      - run:
-          name: Validate front matter
-          command: make validate-front-matter
-
```
marians, line 22: This is moving to a github action.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,32 @@
...
+        if: success()
+        run: pip3 install colored
+      
+      - name: Get list of changed files
+        id: fileList
+        if: success()
+        env:
+          URL: https://api.github.com/repos/${{ github.repository }}/pulls/${{ steps.findPr.outputs.pr }}/files
+        run: |
+          curl --fail -sS $URL | jq -r .[].filename | sort -u > files.txt
```
ubergesundheit, line 27: Have you tried `git ls-tree -r --name-only ${{ github.ref }}`?

ubergesundheit (approved):
LGTM!

Changes seem to work. (Only looked at the output of the github action)

The result output is maybe a little inconvenient to read. Not a blocker, but I think the output for only checking the files changed in a PR, it would be easier to read if it would list files and all problems in the file like this

```
path/to/file/1.md
- WARN: MISSING_FIELD
- FAIL: WRONG_SPELLING

path/to/file/2.md
- FAIL: WRONG_SPELLING
```

.github/workflows/validate.yaml
marians: No, I haven't

marians: It seems to result in everything: https://github.com/giantswarm/docs/runs/2895990635

marians:
I agree the output structure isn't ideal, but didn't want to do too much in one step. I'd rather invest time in commit annotations, which should be more accessible even.

.github/workflows/validate.yaml
ubergesundheit: :sweat_smile: oops yes noticed. Let me quickly invest a minute .. I think it should be possible to have this in a one-liner

ubergesundheit:
>  I'd rather invest time in commit annotations, which should be more accessible even.

Agree! Although I think this is not as easy.. 

> Checks are different from statuses in that they provide line annotations, more detailed messaging, and are only available for use with GitHub Apps.

[Source](https://docs.github.com/en/github/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#types-of-status-checks-on-github)

.github/workflows/validate.yaml
marians: That's for another PR then. I spent too much time on that already.

marians:
With https://github.com/yuzutech/annotations-action it looks relatively simple, but I haven't tried it out yet.

.github/workflows/validate.yaml
ubergesundheit: `git diff --name-only {{ github.base_ref }}` should work. I'll follow up if you want to merge

PR #990, opened 2021-06-23
Use git diff in frontmatter validation workflow

ubergesundheit:
@marians please review ;)

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -18,6 +18,6 @@ jobs:
 
       - name: Validate front matter for changed files
         run: |
-          git fetch origin "${GITHUB_BASE_REF}":"${GITHUB_BASE_REF}"
+          git fetch --no-tags origin "${GITHUB_BASE_REF}":"${GITHUB_BASE_REF}"
           git diff --name-only "refs/heads/${GITHUB_BASE_REF}" -- . | tee files.txt
           cat files.txt | python3 scripts/validate-front-matter/script.py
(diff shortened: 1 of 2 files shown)


.github/workflows/validate.yaml
```diff
@@ -1,33 +1,23 @@
...
         run: pip3 install colored
-            
-      - name: Get list of changed files
-        id: fileList
-        if: success()
-        env:
-          URL: https://api.github.com/repos/${{ github.repository }}/pulls/${{ steps.findPr.outputs.pr }}/files
-        run: |
-          curl --fail -sS $URL | jq -r .[].filename | sort -u > files.txt
-          cat files.txt
```
marians, line 28: Can we keep this step, printing the list to stdout once?

.github/workflows/validate.yaml
```diff
@@ -1,33 +1,23 @@
 name: Validate
 on:
-  push:
-    branches-ignore:
+  pull_request:
```
marians, line 3: This is so strange. From docs I would have never considered `pull_request` as the valid choice for "whenever a push happens to the branch".

The PR details being present simplifies a lot :)

marians (approved):

.github/workflows/validate.yaml
ubergesundheit: Oops, I had enabled automatic merge on approval.

The list is not printed in a separate step, but `tee` will print to stdout and write to file in a singe call. 

I can separate these if you want in a fresh PR

PR #991, opened 2021-06-23
Annotate commits with errors and warnings from front matter validation

marians:
Towards https://github.com/giantswarm/giantswarm/issues/15542

## TODO before merge

Remove dummy markdown pages

## Description

With this PR we create commit annotations for files that fail the front matter check. Here is how the user would access them:

1. Click `Details` in failed `Validate / Front matter problems` check:

    image

2. See the check results

    image

3. User can click the annotation to see it in context

    image


ubergesundheit (approved):
Looks nice. Have not checked the python changes as I’m on my phone, but seems like it works

One nitpick: Please pin the forked action to a commit

marians:
FYI @giantswarm/sig-docs - see main comment

PR #1006, opened 2021-07-02
Add information regarding the kubectl-gs Docker image

marians:


scripts/update-latest-versions/main.sh
```diff
@@ -12,6 +13,7 @@ docker run --rm --entrypoint /bin/sh quay.io/giantswarm/docs-scriptrunner \
     -c "curl --silent https://api.github.com/repos/giantswarm/gsctl/releases/latest|jq -r .tag_name|tr -d '\n'" \
     > src/layouts/shortcodes/gsctl_version.html
 
+# For kubectl-gs, we have to cut the leading 'v' from the tag
 docker run --rm --entrypoint /bin/sh quay.io/giantswarm/docs-scriptrunner \
-    -c "curl --silent https://api.github.com/repos/giantswarm/kubectl-gs/releases/latest|jq -r .tag_name|tr -d '\n'" \
+    -c "curl --silent https://api.github.com/repos/giantswarm/kubectl-gs/releases/latest|jq -r .tag_name | cut -c 2- | tr -d '\n'" \
```
ubergesundheit, line 18: If you want to keep this simpler, your could use the `-j` flag of `jq`, leave the `v` in and use the [`replace`](https://gohugo.io/functions/replace/) hugo function to remove the `v` where required.

ubergesundheit (approved):
LGTM

MarcelMue (approved):

scripts/update-latest-versions/main.sh
marians: Learned something new again, thanks. But it seems as if cut adds a newline.

PR #1007, opened 2021-07-02
Add weekly highlights for 2021-07-09

marians:


marians:
We skipped last week since there were only few items.

ubergesundheit (approved):

PR #1019, opened 2021-07-09
Add metadata for CRDs used on AWS

marians:
This PR adds metadata which will be visible in the form of tags in the CRD listing. Example (before the change):

image


calvix:
do we want to add this even when we only use the CRD for test purposes only and the customer is not exposed to it?

marians:
@calvix The ones I am adding data for here are used in v20 and appear to any user who tries it. So I think they will be relevant soon enough. Antonia is also adding support for those to kubectl-gs via https://github.com/giantswarm/kubectl-gs/pull/380

calvix:
but v20 is an internal release that should not available for customers and they should not try it :p

we do not run components that are responsible for v20 in other than our installations

marians:
Closed, as these CRDs should rather be hidden

PR #1028, opened 2021-07-15
Add article on shared installations

marians:
Towards https://github.com/giantswarm/giantswarm/issues/18035

This PR adds an article "Getting started on a shared installation" which explains the differences between a normal and a shared installation and explains some prerequisites for the customer.


pipo02mix (approved):
LGTM

PR #1048, opened 2021-07-21
Update hugo to v0.85.0

ubergesundheit:
Towards giantswarm/giantswarm#18150

Built locally and browsed a little in the container output


glitchcrab (approved):
seems sensible

PR #1052, opened 2021-07-26
Fix link to replacement CRD in dprecated CRD details page

marians:
If a deprecated CRD points to a replacement CRD, the link was broken. This is fixed in this PR.


piontec (approved):
I haven't checked, but assuming it works - great! Thanks!

PR #1065, opened 2021-08-04
Fixes limitation reason on dashboards page

JosephSalisbury:
We don't limit grafana access on shared installations because the same Prometheus instance would collect several customer's metrics - as we have one Prometheus server per cluster, one Prometheus instance can't collect metrics for several customers.

The limitation is more that multiple customers metrics would be available from the Grafana instance, which we want to avoid.

### Please remember to


QuentinBisson (approved):

marians:
ACK, thanks

PR #1066, opened 2021-08-04
Update MAPI intro article

marians:
- General review
- Removed the info that access is read-only by default
- Added info on admission controller under "what it is"


JosephSalisbury (approved):
wfm

PR #1070, opened 2021-08-05
`kubectl gs template cluster`: replace "master node" with "control plane node"

marians:
@kuosandys Found one of these legacy terms we just talked about


kuosandys (approved):
Thanks for clarifying this!

PR #1083, opened 2021-08-09
Turn off funky syntax highlighting

marians:
Here, the JSON syntax highlighting created a funky effect:

image

I'm applying `nohighlight` here to turn that off.

### DIFF
--- a/src/content/advanced/labelling-workload-clusters/index.md
+++ b/src/content/advanced/labelling-workload-clusters/index.md
@@ -103,7 +103,7 @@ The newly created cluster will be managed by your team in your upstate office an
 
 You've decided on using label keys `my-org/team` and `my-org/environment` to specify the clusters designation.
 
-```json
+```nohighlight
 PUT /v5/clusters/7g4di/labels/
 {
   "labels": {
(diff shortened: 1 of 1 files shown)


stone-z (approved):
So this sent me down a bit of a rabbit hole because I thought it was too bad we have to lose the highlighting altogether. But it sounds like [others have had this issue](https://softwaremaniacs.org/blog/2012/05/09/http-and-json-in-highlight-js/en/), and it might be supported?

I'd suggest trying the `HTTP` and `yaml` options, but if those still don't work I'm fine to drop the highlight. Alternatively, since it seems to just be the numbers in the HTTP line, maybe just change the sample cluster IDs to be alpha only?

marians:
Neat! I'll try this with the HTTP type.

marians:
Just found out again that we are not using highlight.js any more in docs, but apply HUGO's bundled implementation of [Chroma](https://github.com/alecthomas/chroma) on build.

While it's somehow possible to do HTTP highlighting with Chroma, I can't find an example right now. And since the REST API is deprecated anyway, I don't want to invest more time.

PR #1085, opened 2021-08-11
Add 'make validate-last-reviewed'

marians:
Towards establishing a process for periodic content updates (https://github.com/giantswarm/giantswarm/issues/18329)

This PR adds a target `make validate-last-reviewed` which can be executed manually to create a report of pages which have a `last_review_date` that's longer than one year ago.

Technically, this uses the front matter validation script, but skipping all validations except for the last_review_date check.

Preview:

image


stone-z (approved):
Might consider pinning the python dependencies. Otherwise LGTM

PR #1090, opened 2021-08-13
Update `gsctl scale cluster` docs article to be KVM only

marians:
The `gsctl scale cluster` command docs still contained a lot of explanation that's no longer relevant.

Now that there are almost only clusters with node pools on AWS and Azure, the command does not apply any more on these providers. The command is effectively only usable on KVM now.

I also added a pointer to the right command to use for AWS and Azure (`gsctl update nodepool`).


marians:
Moving on

PR #1101, opened 2021-08-23
Mapi key pairs gude: use sha256sum instead of shasum

marians:
Closes https://github.com/giantswarm/giantswarm/issues/18595

It turns out `shasum` is generally available on Mac OS, but on on most Linux systems it seems.


axbarsan (approved):
Should we specify that on macOS they should use `shasum` instead?

marians:
> Should we specify that on macOS they should use `shasum` instead?

In the Prerequisites section it is mentioned which utilities are needed. I would hope that each user would verify that.

PR #1119, opened 2021-09-14
Fix front matter validation

marians:
This PR fixes a bug in front matter validation ([example](https://github.com/giantswarm/docs/runs/3585202091?check_suite_focus=true))


scripts/validate-front-matter/script.py
```diff
@@ -388,18 +388,22 @@ def validate(content, fpath, validation):
...
     num_lines = len(list(content.split("\n")))
 
     # Detect whether there is front matter
     if not content.endswith('\n'):
-        result.append({
+        result['checks'].append({
             'check': NO_TRAILING_NEWLINE,
             'line': num_lines, # last line
         })
         return result
```
marians, line 404: The actual bug was that here (among others), the returned result had a different structure than expected.

scripts/validate-front-matter/script.py
```diff
@@ -415,14 +419,14 @@ def validate(content, fpath, validation):
         if fmstring != "":
             fm = load(fmstring, Loader=Loader)
     except Exception as e:
-        result.append({
+        result['checks'].append({
             'check': NO_FRONT_MATTER,
             'line': 1,
         })
         return result
```
marians, line 426: ... here as well

ubergesundheit (approved):
LGTM

PR #1126, opened 2021-09-16
Create an automated way of fetching and updating the website content

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18920

This PR adds an automated (80% now, 100% later) way of grabbing and updating the header and footer from the main website.

The automated part:
- We use a headless browser to access `https://www.giantswarm.io/why-giant-swarm`
- We grab the header HTML element (selector `.header-container-wrapper`) and all of its children.
- We go through all the CSS rules on the page and see which of them match the header element or its children and we grab them.
- We update the header HTML partial file.
- We grab the footer HTML element (selector `.footer-container-wrapper`) and all of its children.
- We go through all the CSS rules on the page and see which of them match the footer element or its children and we grab them.
- We update the footer HTML partial file.
- We merge together the CSS rules required to render the header and the footer and we remove duplicates.
- We add the CSS overrides from `/scripts/update-website-content/overrides.css` to the common CSS.
- We write the CSS into the website styles partial file.

CSS overrides are necessary, because without them, the automation gives us something that doesn't look 100% the same as the website. There are specific rules extracted from the website that are overriden by the `docs` and/or `bootstrap`

For the JS part however, I extracted that manually, to not spend 12312 weeks into developing a pattern matching system.

You can run the script by executing `make update-website-content`.

Unless there are big functionality changes to the website header/footer, we can run the automation and update everything once in a while (via CI, for example). That would work without any human intervention.


Makefile
```diff
@@ -108,9 +108,10 @@ linkcheck-internal:
 	docker kill server
 	docker kill linkchecker
 
-grab-main-site-header-footer:
-	curl -s https://www.giantswarm.io/why-giant-swarm | sed -n '/^<!-- BEGIN SITE_HEADER -->/,/^<!-- END SITE_HEADER -->/p;/^<!-- END SITE_HEADER -->/q' | sed '1d;$d' > src/layouts/partials/gs_header.html
-	curl -s https://www.giantswarm.io/why-giant-swarm | sed -n '/^<!-- BEGIN SITE_FOOTER -->/,/^<!-- END SITE_FOOTER -->/p;/^<!-- END SITE_FOOTER -->/q' | sed '1d;$d' > src/layouts/partials/gs_footer.html
+update-website-content:
+	cd ./scripts/update-website-content && \
```
axbarsan, line 112: Expects you to have Node.js installed

src/layouts/partials/styles.html
```diff
@@ -3,7 +3,8 @@
 {{- $cssOpts      := (dict "targetPath" $cssTarget "outputStyle" "compressed") }}
 {{- $css := resources.Get $sass | toCSS $cssOpts | minify | fingerprint }}
 
-<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto+Slab:300,700|Roboto:300,700,300italic|Inconsolata:400,700">
+<link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto+Slab:300,700|Roboto:300,500,700,300italic|Inconsolata:400,700">
 <link rel="stylesheet" type="text/css" href="https://s.giantswarm.io/giantswarmio/0.10.18/1/bootstrap.min.css">
+<link rel="stylesheet" type="text/css" href="https://cdn2.hubspot.net/hubfs/273774/fontawesome/v5/latest.css">
```
axbarsan, line 8: Necessary for all the new icons. But since this is shared between the website and the docs, it means that caching helps us, so the impact is not that big.

marians (approved):
Great as far as I can tell :) Didn't test.

kuosandys (approved):
This is really cool!

PR #1127, opened 2021-09-16
Replace --owner with --organization in `kubectl gs template` commands

marians:
Follow-up for https://github.com/giantswarm/kubectl-gs/pull/471


pipo02mix (approved):
lovely, I scratch this out from my TODO list

PR #1178, opened 2021-10-26
Add article explaining how to provide cloud provider credentials via the Management API

marians:
Towards https://github.com/giantswarm/roadmap/issues/480

So far, only AWS is addressed.


marians:
Dear people, I've almost rewritten the article after understanding better how the credentials handling works. Here is the outline of the current structure:

image

cluster.yaml
```diff
@@ -0,0 +1,94 @@
+apiVersion: cluster.x-k8s.io/v1alpha3
```
axbarsan, line 1: I guess these 2 files have been committed by accident

marians: You're right! Thanks

marians: I removed them. FYI: The secret.yaml only contained dummy data.

PR #1200, opened 2021-11-11
Add weekly highlights for the week ending on 2021-11-12

marians:
Please add your input until Friday 14:00 CET


src/content/changes/highlights/2021-11-12.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Management API
+
+For `kubectl get apps.application.giantswarm.io` (or just `kubectl get apps`) we changed the output columns for the App CRD. With that, for apps installed in workload clusters, you can now differentiate the installed version from the desired version.
+
+In automation, please make sure to use JSON or YAML output of `kubectl get` commands instead if plain text, so your implementation is not affected by a change like the above.
+
+## Monitoring
+
+If you are using Flux CD in the management cluster for your own good, you'll want to check out the two new Grafana dashboards on Flux CD metrics which we added this week.
```
kubasobon, line 28:
```suggestion
If you are using Flux CD in the management cluster for your own good, you'll want to check out the two new Grafana dashboards on Flux CD metrics which we added this week. `Flux Cluster Stats` showcases reconcilers and operator data, while `Flux Control Plane` will give you an insight into resource usage, number of API requests, and ongoing operation rates.
```

marians:
Please, @giantswarm/sig-product , don't be shy!

src/content/changes/highlights/2021-11-12.md
```diff
@@ -0,0 +1,32 @@
...
+title: Highlights for the week ending November 12, 2021
+changes_categories:
+  - Highlights
+owner:
+  - https://github.com/orgs/giantswarm/teams/sig-product
+---
+
+## Apps
+
+[Flux](https://github.com/giantswarm/flux-app) v0.7.0 provides the flux toolkit version 0.21.0 and adds service discovery labels for Prometheus, so that metrics are scraped.
```
puja108, line 12:
```suggestion
[Flux](https://github.com/giantswarm/flux-app) v0.7.0 provides the Flux Toolkit version 0.21.0 and adds service discovery labels for Prometheus, so that metrics are scraped.
```

this is a proper noun as used in upstream website/docs

src/content/changes/highlights/2021-11-12.md
```diff
@@ -0,0 +1,32 @@
...
+
+## User interfaces
+
+In the web UI, we now display if a cluster has upgrades scheduled.
+
+`kubectl gs login` has been improved in several ways. Most importantly, the command now also supports using a service account token for authentication via the `--token` flag. This facilitates creation of workload cluster client certificates especially in automation. ([Docs](https://docs.giantswarm.io/ui-api/kubectl-gs/login/))
+
+Additionally, when creating a client certificate for a workload cluster, the flag `--organization` is now effectively optional.
```
puja108, line 20:
```suggestion
Additionally, when creating a client certificate for a workload cluster, the flag `--organization` is now optional.
```

or what do you mean by "effectively", is there any case where it isn't?

marians: Yeah. In the theoretical case that there are two clusters with the same name accessible to this user (which may be possible with CAPI), there will be an error and the user will have to specify the `--organization` flag in addition.

puja108:
```suggestion
Additionally, when creating a client certificate for a workload cluster, the flag `--organization` is now optional as long as the workload cluster name is unique.
```

maybe write it out here then as it's just a few more words.

Also,

IIRC we were validating to only allow unique names even if CAPI would allow this.

To be clear, I'd be ok with removing our limitation, but not sure that is possible, as we have relied on the unique `name` before and those places would then need to rely on `namespace/name`. We'd need to see if we can make that work with our already implemented components working with future CAPI stuff /cc @alex-dabija @cornelius-keller mainly, but might even be influenced by things honeybadger (APO) and atlas (PMO) have built.

marians: OK, let's take this to KaaS.

This PR here is the addition of an announcement that has been posted on Friday already. I think there isn't much harm in adding it to our "archive" in docs this way. But I still don't have an approval.

marians: See https://gigantic.slack.com/archives/CNX3Z835G/p1636977113011700

puja108 (approved):

PR #1213, opened 2021-11-18
Take into account unsupported providers when generating release changelogs

axbarsan:
Encountering an unsupported provider (e.g. OpenStack) when generating release changelogs makes the script crash. This prevented the `Update generated content` GitHub action from doing its thing.

### DIFF
--- a/scripts/aggregate-changelogs/script.py
+++ b/scripts/aggregate-changelogs/script.py
@@ -35,7 +35,7 @@ def get_releases(client, repo_shortname):
         # skip unpublished releases
         if release.published_at is None:
             continue
-        
+
         body = ""
         if release.body is not None:
             body = link_pull_requests(release.body, repo_shortname)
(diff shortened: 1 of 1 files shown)


scripts/aggregate-changelogs/script.py
```diff
@@ -116,7 +116,9 @@ def parse_release_yaml(data, repo_shortname, provider, relative_path):
         raise Exception('Unexpected release data format. No ".spec" attribute found.')
     if 'date' not in data['spec']:
         raise Exception('Unexpected release data format. No ".spec.date" attribute found.')
-    
+    if provider is None:
+        raise Exception('Unexpected provider.')
```
axbarsan, line 120: It can be `None` when it's not `aws`, `azure` or `kvm`

scripts/aggregate-changelogs/script.py
```diff
@@ -144,43 +146,43 @@ def get_cluster_releases(repo_shortname):
         for root, _, files in os.walk(tmpdir):
             if os.path.basename(root) in ('aws', 'azure', 'kvm'):
                 provider = os.path.basename(root)
```
axbarsan, line 148: We only support `aws`/`azure`/`kvm` for generating changelogs

marians (approved):
LGTM if it works

PR #1245, opened 2021-12-09
Update CRD docs generator and configuration

marians:
Towards https://github.com/giantswarm/roadmap/issues/616

This PR

- uses a modified crd-docs-generator v0.8.0, which pulls CRDs from several repositories
- moves CRD metadata (back) here to the docs repository, as the metadata is only maintained for docs and it's better to have it in one place here than spread over the operator repositories.

When reviewing, you can look at the separate commits for more clarity. One commit contains the resulting changes to generated CRD docs pages.

The fact that for some cluster-related CRDs we are adding version v1alpha2 but removing v1alpha4 is not caused by this PR, but instead is the result of some CRD change in giantswarm/apiextensions.

### DIFF
--- a/scripts/update-crd-reference/crd.template
+++ b/scripts/update-crd-reference/crd.template
@@ -137,7 +137,7 @@ This CRD is being replaced by <a href="../{{ .FullName }}/">{{ .ShortName }}</a>
 {{ range $versionSchema.Annotations }}
 <div class="annotation">
 <div class="annotation-header">
-<h3 class="annotation-path" id="{{.APIVersion}}-{{.Annotation}}">{{.Annotation}}</h3>
+<h3 class="annotation-path" id="{{.CRDVersion}}-{{.Annotation}}">{{.Annotation}}</h3>
 </div>
 <div class="annotation-body">
 <div class="annotation-meta">
--- a/scripts/update-crd-reference/main.sh
+++ b/scripts/update-crd-reference/main.sh
@@ -2,21 +2,15 @@
 
 set -e
 
-CRD_DOCS_GENERATOR_VERSION=0.7.1
+CRD_DOCS_GENERATOR_VERSION=0.8.0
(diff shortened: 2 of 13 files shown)


scripts/update-crd-reference/crd.template
```diff
@@ -137,7 +137,7 @@ This CRD is being replaced by <a href="../{{ .FullName }}/">{{ .ShortName }}</a>
 {{ range $versionSchema.Annotations }}
 <div class="annotation">
 <div class="annotation-header">
-<h3 class="annotation-path" id="{{.APIVersion}}-{{.Annotation}}">{{.Annotation}}</h3>
+<h3 class="annotation-path" id="{{.CRDVersion}}-{{.Annotation}}">{{.Annotation}}</h3>
```
marians, line 140: A field name change that comes with crd-docs-generator v0.8.0

scripts/update-crd-reference/main.sh
```diff
@@ -2,21 +2,15 @@
...
 DESTINATION=src/content/ui-api/management-api/crd
 
 # Clear output folder
 find ${DESTINATION} -type f -not -name "_index.md" | xargs rm
 
-# Determine apiextensions version
-VERSION=$(docker run --rm --entrypoint /bin/sh quay.io/giantswarm/docs-scriptrunner -c "curl --silent https://api.github.com/repos/giantswarm/apiextensions/releases/latest|jq -r .tag_name|tr -d '\n'")
-
-echo "Version: ${VERSION}"
-
```
marians, line 15: We no longer get the latest version in this script. This is now hard coded in the config.

We can find ways in the future to keep that up-to-date.

stone-z (approved):
Seems ok to me. That config is going to be fun to maintain, I don't see a way around it for the time being though

marians:
Thanks for checking @tfussell !

PR #1264, opened 2022-01-07
Creating an app catalog

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/19941

### Please remember to

- [x] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/main/CONTRIBUTING.md#front-matter) associated with any updated docs


src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+
+Create the git repository in GitHub and enable [GitHub Pages](https://docs.github.com/en/pages/quickstart)
+for the `main` branch.
+
+Initialize the empty helm index.yaml with `helm repo index .` and commit it to
+the repository.
+
+## Pushing an app to an app catalog
+
+TODO
```
rossf7, line 51: @piontec Just to confirm. 😅 Do you agree we should use ABS for this? 

PTAL at the questions below.

piontec: Yes, definitely. We should make it our 'standard' for building charts.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+  - How can I push an app to my app catalog?
+---
+
+# Creating an app catalog
+
+## Overview
+
+An app catalog is a collection of apps that can be deployed using the Giant Swarm App Platform.
+We fully support Helm and an app catalog is also a Helm [Chart Repository](https://helm.sh/docs/topics/chart_repository/).
+Each app catalog has its own [catalog]({{< relref "/ui-api/management-api/crd/catalogs.application.giantswarm.io.md" >}})
```
piontec, line 25: Should we make it capital 'C' - `Catalog`, like the full CR name?

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+metadata extensions will work.
+
+But I have some questions / concerns.
+
+- Should we link to the ABS [tutorial](https://github.com/giantswarm/app-build-suite/blob/master/docs/tutorial.md)?
+- Or should we create another docs page and link them?
+- ABS doesn't commit the files to the repo (which is intentional) but I think we need a solution for that.
+- We could suggest architect-orb but that feels weird since its mainly internal and customer may not be using
+Circle CI.
+- Should we create a simple GitHub Action for this?
```
piontec, line 63: Sharing architect(-orb) is not a good idea, IMHO. There are already github actions for simple commits like this, I'm using one for creating PRs here: https://github.com/giantswarm/github/blob/7701fdadf34ce91d29fa370ef78546bf46c3ae2d/.github/workflows/update_pre_commit.yaml#L23
I'm sure we can find something ready for commits as well.
But that's the right direction overall, I think.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+  namespace: 2hr7z
+spec:
+  catalog: example-catalog
+  catalogNamespace: org-example
+  kubeConfig:
+    inCluster: false
+  name: example-app
+  namespace: default
+  version: 0.1.0
+```
```
piontec, line 128: Should we also include info about how `ace`s are generate?

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+  - How can I create a catalog CR for my app catalog?
+  - How can I push an app to my app catalog?
+---
+
+# Creating an app catalog
+
+## Overview
+
+An app catalog is a collection of apps that can be deployed using the Giant Swarm App Platform.
+We fully support Helm and an app catalog is also a Helm [Chart Repository](https://helm.sh/docs/topics/chart_repository/).
```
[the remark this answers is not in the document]
rossf7: This is what we have in the overview page but I agree it could be clearer. I'll take another look in the next revision.

src/content/app-platform/create-catalog/index.md
rossf7: I lose track of what style we use :) I'll check 👍

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,128 @@
...
+## Pushing an app to an app catalog
+
+TODO
+
+I think we should recommend using ABS (app-build-suite) and this also means our
+metadata extensions will work.
+
+But I have some questions / concerns.
+
+- Should we link to the ABS [tutorial](https://github.com/giantswarm/app-build-suite/blob/master/docs/tutorial.md)?
```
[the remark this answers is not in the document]
rossf7: I'll start with linking to the tutorial but it feels a bit weird. I'd expect at least a getting started page in docs.

src/content/app-platform/create-catalog/index.md
rossf7: Agree architect / orb doesn't work. I like the idea of a simple github action. If customers are using other options for CI the action still works as an example.

src/content/app-platform/create-catalog/index.md
rossf7: Good point. I'll add.

rossf7 (commented):
@piontec Thanks for the feedback that unblocks this. I'll do another draft but it's likely to be next week.

src/content/app-platform/create-catalog/index.md
rossf7: Done. Last section explains this and shows listing ACE CRs for catalog with kubectl.

src/content/app-platform/create-catalog/index.md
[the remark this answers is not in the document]
rossf7: Replaced support with extend.

src/content/app-platform/create-catalog/index.md
rossf7: We use caps. I've updated.

rossf7 (commented):
Doc is updated. Pushing apps to the catalog section is still WIP

MarcelMue (approved):
Works for me!

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,161 @@
...
+name: 'Push to App Catalog'
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  push_to_app_catalog:
+    uses: giantswarm/app-build-suite/.github/workflows/push-to-app-catalog.yaml@master
```
rossf7, line 72: How should we version the action?

I think we should use tagged releases and recommend customers set up dependabot to keep the action updated. WDYT?

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot

piontec (approved):
LGTM

puja108 (commented):
In general I have no issue with this doc and this topic is also needed, however, I wonder if the use case is always "creating a helm repo". To me you are describing the creation and management of a helm repository here (in a very opinionated way with git and our tooling) and only in the end actual create a catalog.

In my mind the use case would be more like:
1. I want to install some apps from helm charts -> create a catalog
2. Do you already have a helm repository or do you need help creating/hosting one?
2a. I already have a helm repo -> jump to 3.
2b. I don't have a repo -> how to create a repo with git
3. create catalog CR

I say this because we get inquiries from customers if we could host an upstream chart for them in our catalogs. The most straight forward docs for that enquiry would IMO be "How to create a catalog for an upstream/existing helm repo".

WDYT?

rossf7:
@puja108 Good point the doc only covers creating your own catalog and not using community helm repos.

I think it can be restructured to support both like you suggest. We could also point customers to [artifacthub.io](https://artifacthub.io/) as a source for finding community repos.

However some community repos won't work due to https://github.com/giantswarm/giantswarm/issues/20285

This is a problem with our initial implementation of app platform. We generate the tarball URL for use in chart CRs instead of getting it from the `index.yaml` of the helm repo. We plan to fix this soon and we could add a note to the docs until then. WDYT?

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+to keep the GitHub Action up to date.
+
+## Create Catalog CR
+
+Once you've created your app catalog and published some apps you need to create
+a Catalog CR in the management cluster to register it with app platform. This
+can be done using [kubectl gs template catalog]({{< relref "/ui-api/kubectl-gs/template-catalog" >}}).
+
+You should create the Catalog CR in the [organization]({{< relref "/general/organizations" >}})
+namespace where the apps will be used.
```
marians, line 95: Any alternative, if the catalog is supposed to be used by more than one org? Use the default namespace?

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+To configure the action you need to add a secret called `PAT` to the app's git
+repository. This must contain a [Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
+that has write permission to the app catalog git repository.
+
+We recommend using [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot)
+to keep the GitHub Action up to date.
+
+## Create Catalog CR
```
marians, line 88: While this headline is not wrong, can we make it more goal-focused? Creating the CR is a means to an end. I think "Register the catalog" or something like that would work better.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+spec:
+  description: An example Catalog
+  logoURL: https://example.com/logos/example-logo.png
+  storage:
+    URL: https://example.github.io/example-catalog/
+    type: helm
+  title: example
+```
+
+## Referencing an app catalog in an App CR
```
marians, line 122: Again, the headline should be more goal or use case oriented. In this case, I think it means "Installing an app from the catalog to a workload cluster", right?

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+  catalog: example
+  catalogNamespace: org-example
+  kubeConfig:
+    inCluster: false
+  name: example-app
+  namespace: default
+  version: 0.1.0
+```
+
+## Viewing app metadata
```
marians, line 154: I would call this something like "Browsing the catalog" or "Inspecting the catalog"

marians (approved):
Thanks @rossf7 for adding this page and filling a gap we've had for some time! I think the quality is pretty high and it could be merged as is.

My comments are all suggestions for further improvements and nit-picks.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+There are [multiple options](https://helm.sh/docs/topics/chart_repository/#hosting-chart-repositories)
+for serving the catalog over HTTP including GitHub Pages or tools like Harbor or ChartMuseum which run in a
+Kubernetes cluster. At Giant Swarm we use GitHub Pages and this is what we will
+cover in this guide.
+
+## Creating an app catalog hosted using GitHub Pages
+
+First, you should choose a name for your catalog. We recommend using the suffix
+`-catalog` to make it clear this Git repository hosts an app catalog.
```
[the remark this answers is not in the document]
rossf7: Yes any name will work. Good to clarify that.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+Create the Git repository in GitHub as a public repository and enable
+[GitHub Pages](https://docs.github.com/en/pages/quickstart) for the `main` branch.
+
+## Publishing an app to the app catalog
+
+Now you can configure your apps to be published to your catalog. We have created
+a GitHub Action for `app-build-suite` that will automate this.
```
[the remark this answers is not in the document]
rossf7: Will do 👍

src/content/app-platform/create-catalog/index.md
rossf7: Yes register is better 👍

src/content/app-platform/create-catalog/index.md
rossf7: Yes, default would be the alternative if the catalog is needed for multiple orgs and customers should have permission. I'll update this section.

src/content/app-platform/create-catalog/index.md
rossf7: Yes, this is for installing the app.

src/content/app-platform/create-catalog/index.md
rossf7: Thanks I like inspecting for this.

rossf7 (commented):
Changes made. PTAL if you have any final feedback.

src/content/app-platform/create-catalog/index.md
```diff
@@ -0,0 +1,183 @@
...
+  - How can I inspect the apps in an app catalog?
+---
+
+# Creating an app catalog
+
+## Overview
+
+An app catalog is a collection of apps that can be deployed using the Giant Swarm App Platform.
+We extend Helm and an app catalog is also a Helm [<span class="x x-first x-last">chart repository</span>](https://helm.sh/docs/topics/chart_repository/).
+Each app catalog has its own [<span class="x x-first x-last">`</span>Catalog<span class="x x-first x-last">`</span>]({{&lt; relref "/ui-api/management-api/crd/catalogs.application.giantswarm.io.md" &gt;}})
```
rossf7, line 28:
```suggestion
Each app catalog has its own [<span class="x x-first x-last">`</span>Catalog<span class="x x-first x-last">`</span>]({{< relref "/ui-api/management-api/crd/catalogs.application.giantswarm.io.md" >}})
```

marians:
Why do we have this `<span class="x x-first x-last">` formatting?

rossf7:
> Why do we have this <span class="x x-first x-last"> formatting?

@marians Hmm weird it was in your suggested edits see https://github.com/giantswarm/docs/pull/1264/commits/1339728e9e80121490c14fa337879cac05a150bb but I don't see them now.

I think it could have been a github glitch. Shall I remove them?

PR #1286, opened 2022-01-31
kubectl gs login: Update content for workload cluster client certs, explain certificate X.509 fields

marians:
Towards https://github.com/giantswarm/roadmap/issues/770

### Please remember to

- [ ] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/main/CONTRIBUTING.md#front-matter) associated with any updated docs

### DIFF
--- a/src/content/ui-api/kubectl-gs/login.md
+++ b/src/content/ui-api/kubectl-gs/login.md
@@ -132,8 +132,6 @@ With these values defaulting to random strings, we ensure that a client presenti
 
 To assign specific group memberships, or in other terms, to specify additional `O` values in the created certificate, you'll have to use the `--certificate-group` flag when executing the command. This flag can be set multiple times, with one group name each, as shown in the example below.
 
-If `--certificate-group` flag, you can still create a valid certificate, however the client presenting the certificate would not get any permissions in the Kubernetes API. The reason is that without that flag, both the `CN` attribute (interpreted as the user name by the Kubernetes API) and the `O` attribute of the certificate (interpreted as group name) will contain random values, so that no role bindings can match these names.
-
 Example command:
 
 ```nohighlight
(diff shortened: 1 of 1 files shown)


marians:
Hi @giantswarm/team-rainbow ! The discovery that a user cannot specify the CN of a client certificate made me want to explain it better at least. As a consequence, I re-organized and extended our docs about workload cluster login.

src/content/ui-api/kubectl-gs/login.md
```diff
@@ -109,15 +109,66 @@ kubectl gs login example \
...
+```
+
+When presenting such client certificate to the Kubernetes API server, the `CN` field is interpreted as the user identifier, while the `O` field is interpreted as a list of groups the user is a member of.
+
+With these values defaulting to random strings, we ensure that a client presenting this certificate does not accidentally acquire any permissions based on pre-defined group or user names in any (cluster) role bindings.
+
+To assign specific group memberships, or in other terms, to specify additional `O` values in the created certificate, you'll have to use the `--certificate-group` flag when executing the command. This flag can be set multiple times, with one group name each, as shown in the example below.
+
+If `--certificate-group` flag, you can still create a valid certificate, however the client presenting the certificate would not get any permissions in the Kubernetes API. The reason is that without that flag, both the `CN` attribute (interpreted as the user name by the Kubernetes API) and the `O` attribute of the certificate (interpreted as group name) will contain random values, so that no role bindings can match these names.
```
kuosandys, line 135: I was wondering if the start of the sentence could be a little more specific: "If `--certificate-group` flag" is a bit unclear for me

kuosandys:
Thanks for updating the content and adding extra details 👍 Just the small suggestion above, otherwise looks good (especially the installation name in the example)!

src/content/ui-api/kubectl-gs/login.md
marians: It's definitely bad language :o. Anyway, I basically forgot to adapt that whole paragraph in line 135. Will do.

marians:
Removed that left-over sentence

kuosandys (approved):

PR #1295, opened 2022-02-04
kubectl-gs login: update docs for --keep-context and --self-contained flags

kuosandys:
Update for kubectl-gs `login` documentation for the changes: https://github.com/giantswarm/kubectl-gs/pull/660 and https://github.com/giantswarm/kubectl-gs/pull/661.


anvddriesch (approved):

src/content/ui-api/kubectl-gs/login.md
```diff
@@ -56,6 +56,8 @@ For **management cluster** authentication, the following flags are available:
 
 - `--token`: Use a Bearer token instead of OIDC authentication, e. g. a service account token.
 
+- `--keep-context`: Keep the current kubectl context selected after authenticating. If this flag is not set or false, the newly created context will be selected as the current-context.
```
marians, line 59: This is placed in the `**management cluster**` section. Is that correct? Is the flag only applicable with MC login, not WC?

kuosandys: Yes it's only applicable with MC login at the moment, but I just saw that there are some changes in https://github.com/giantswarm/kubectl-gs/pull/661/files that will make it available for WC as well.

I think maybe this PR is premature - we can also align the docs after the other PR is also merged?

marians: Yes, we have to wait for a release anyway.

marians (approved):

PR #1298, opened 2022-02-04
Add docs related to CAPZ cluster creation using v20.0.0-alpha1

fiunchinho:
### Please remember to

- [X] Give the PR a meaningful title -- it will be shown in the release notes
- [ ] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/main/CONTRIBUTING.md#front-matter) associated with any updated docs

PR #1300, opened 2022-02-07
Use crd-docs-generator v0.9.0

marians:
To be merged into https://github.com/giantswarm/docs/pull/1298

As a result, there are some additions, but no removals of CRDs, as far as I can see. Please check whether this meets your expectations.


fiunchinho (approved):
that looks much better, thank you!

PR #1308, opened 2022-02-09
Dex article: add `teamNameField: slug` to example github connector

marians:
We found that the github connector for Dex defaults to the team name if there is no `teamNameField` setting in the connector config. As the team name can contain spaces and uppercase letters, this is brittle. Using the team slug (the part that appears in the team's URL) is more robust and hence `teamNameField: slug` is our recommended setting.


anvddriesch (approved):
Will people know what is meant when they read `slug`? Because I didnt'.

T-Kukawka (approved):

T-Kukawka:
Maybe adding a comment after slug with a small explanation? It is upstream 'naming'

PR #1313, opened 2022-02-11
Add content for Rainbow

marians:


gianfranco-l (commented):
kubectl gs is missing url or the link is not supposed to be there? @marians

PR #1320, opened 2022-02-16
Update 'template' intros

marians:
Goal of this PR is to make it clear where the output of the kubectl-gs `template <noun>` commands is to be applied.


LolloneS (approved):
Looks great! Thank you :)

PR #1329, opened 2022-02-24
Update CI configuration

ubergesundheit:
This PR aims to replace outdated and failing CI configuration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,8 @@ jobs:
             container_tag=$(architect project version)
             echo "$container_tag"
             docker run -d --rm --name docs -p 8080:8080 quay.io/giantswarm/docs:$container_tag
-            CURL_OUTPUT=$(curl --no-buffer -s http://localhost:8080)
+
+            CURL_OUTPUT=$(docker run --link docs quay.io/giantswarm/curl:7.81.0 curl --no-buffer http://docs:8080)
             echo $CURL_OUTPUT | grep --quiet "Giant Swarm"
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,18 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,46 +1,40 @@
...
+            docker run -d --rm --name docs -p 8080:8080 quay.io/giantswarm/docs:$container_tag
 
-      - deploy:
-          name: Deploy (only if in default branch)
-          command: |
-            echo "Deploy only if '${CIRCLE_BRANCH}' == 'main'"
-            if [ "${CIRCLE_BRANCH}" == "main" ]; then
-              ./architect deploy
-            fi
+            CURL_OUTPUT=$(docker run --link docs quay.io/giantswarm/curl:7.81.0 curl --no-buffer http://docs:8080)
```
ubergesundheit, line 36: Since we switched away from `machine` based building, any container started with `docker run` is not accessible using `curl localhost`. Here we just spin up a second container linked to the first to execute curl

helm/docs-app/templates/deployment.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       containers:
         - name: {{ .Values.image.name }}
           # Public image required
-          image: quay.io/giantswarm/{{ .Values.image.name }}:{{ .Values.image.sha }}
+          image: quay.io/giantswarm/{{ .Values.image.name }}:{{ .Values.image.tag }}
```
ubergesundheit, line 25: Once merged, container tags are constructed with `<git-tag>-<git-sha>`

.circleci/config.yml
ubergesundheit, line 36: Since we switched away from `machine` based building, any container started with `docker run` is not accessible using `curl localhost`. Here we just spin up a second container linked to the first to execute curl

marians (approved):
Thänks!

PR #1357, opened 2022-03-09
Set HUGO_DOCS_VERSION environment variable on build time

marians:
Towards publishing the docs release version in a content meta tag.

When setting the variable as an env variable in the pod, I missed that it has to be set on build time instead.


ubergesundheit (approved):
LGTM

PR #1386, opened 2022-03-31
Use Organization CRD from organization-operator

marians:
As the organization-operator repo is now the canonical source of the Organization CRD, this PR changes the CRD docs to use the new source.


marians:
FYI @fiunchinho

PR #1392, opened 2022-04-08
Add PDB and anti affinity

marians:
Towards https://github.com/giantswarm/giantswarm/issues/21641

### DIFF
--- a/helm/docs-app/templates/pdb.yaml
+++ b/helm/docs-app/templates/pdb.yaml
@@ -1,0 +1,11 @@
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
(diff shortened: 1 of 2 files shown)


kuosandys (approved):

helm/docs-app/templates/pdb.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "resource.dex.name" . }}
```
MarcelMue, line 5: It seems like this isn't working - helm fails to find `resource.dex.name`

marians: Yeah, that was a dull mistake

PR #1432, opened 2022-05-11
Add CSV export script

marians:
This PR adds a Python script to export a list of articles/pages as CSV. I needed this for internal purposes.

It's executed by `make export-csv` and results in a file `export.csv` being written into the main folder.


ubergesundheit (approved):
👍

PR #1443, opened 2022-05-18
Add script for Google Analytics tracking, active if consent is given

marians:
Towards https://github.com/giantswarm/giantswarm/issues/22128

This PR adds JavaScript code to check if the user has given consent based on the Hubspot GDPR banner. If yes, we add Google Analytics tracking code to the DOM.

The GA4 property is one exclusively created for docs.


marians:
It turned out that this is not ready as it is. We want to use a different GA4 property for the entire domain, but that isn't set up yet.

pipo02mix:
what we do with this @marians ?

pipo02mix:
I am closing this for now

PR #1444, opened 2022-05-18
Dissolve reference section

marians:
Towards https://github.com/giantswarm/giantswarm/issues/20188

This PR moved the three pages currently contained in the "Reference" top level section into other sections.

We want to make room on the top level and the reference section was one that didn't fulfil a real purpose.


marians:
@giantswarm/team-rainbow As SIG Docs seems to be slacking, maybe you can help me out with a review? <3

kuosandys (approved):
Looks good to me 👍

kuosandys:
Btw "Reference" still shows up in the side bar - just checking that we still want that to be there, right?

marians:
No, good catch! I had missed that, also with the Kubernetes basics section.

PR #1504, opened 2022-07-19
Add docs on Ingresses/TLS

LolloneS:
First round, please give suggestions when you're back @Gacko 😄 

### Please remember to

- [x] Add (or remove) [user questions](https://github.com/giantswarm/docs/blob/main/CONTRIBUTING.md#front-matter) associated with any updated docs

### DIFF
--- a/src/content/advanced/ingress/multi-nginx-ic/index.md
+++ b/src/content/advanced/ingress/multi-nginx-ic/index.md
@@ -16,7 +16,7 @@ user_questions:
   - How do I configure NGINX Ingress Controller for internal traffic?
   - How do I override the NodePorts on KVM Ingresses?
   - How do I configure NGINX Ingress Controller to allow weak ciphers?
-last_review_date: 2021-09-01
+last_review_date: 2022-07-19
 ---
 
 NGINX ingress controller handles [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources, routing traffic from outside the Kubernetes cluster to services within the cluster.
@@ -112,6 +112,20 @@ Similarly, cloud load balancer created for each NGINX installation on AWS and Az
 For NGINX IC running on on-prem (KVM) workload clusters there's no out-of-the-box `LoadBalancer` Service type support. Therefore, NGINX Service type defaults to `NodePort`. For every NGINX IC App installation, one must assign a set of unique http and https node ports. The default NGINX http and https node ports are `30010` and `30011`. The example sets `31010` and `31011` as overrides for the internal NGINX.
 
 More information on this topic can be found in document [Services of type LoadBalancer]({{< relref "/content/advanced/ingress/service-type-loadbalancer/index.md" >}}).
+
+It is also possible to only install a single Nginx Ingress Controller App and to delegate both internal and external traffic to it. Here is a minimal working example on how to achieve this goal.
+
+```yaml
+controller:
+  service:
+    public: true  # default value
+    subdomain: "ingress"  # default value
+    internal: 
+      enabled: true  # default is `false`
+      subdomain: "ingress-internal"  # default value
+```
+
+In other words, it is sufficient to set `controller.service.internal.enabled` to `true` to create two services: one for public traffic and one for private one. On cloud providers, the Services we create will be of type `LoadBalancer`; on premise, depending on the platform, they might be either of type `LoadBalancer` or `NodePort`.
 
 ## Using weak ciphers for legacy clients
 
--- a/src/content/advanced/tls-certificates/index.md
+++ b/src/content/advanced/tls-certificates/index.md
@@ -1,0 +1,103 @@
+---
+linkTitle: TLS certificates with cert-manager
+title: Obtaining TLS certificates for Ingresses with cert-manager
+description: Configure cert-manager to automatically obtain TLS certificates for Ingresses
+weight: 20
+owner:
+  - https://github.com/orgs/giantswarm/teams/team-cabbage
+user_questions:
+  - How do I install cert-manager?
+  - How do I obtain TLS certificates?
+last_review_date: 2022-07-19
+---
+
+Exposing HTTP services using an Ingress is a pretty straightforward task in Kubernetes. Using plain HTTP is, however, discouraged in favor of using HTTPS. `cert-manager` provides a simple, declarative way to automatically obtain TLS certificates for the endpoints we expose.
+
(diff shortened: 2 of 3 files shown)


src/content/advanced/tls-certificates/index.md
```diff
@@ -0,0 +1,104 @@
...
+  - https://github.com/orgs/giantswarm/teams/team-cabbage
+user_questions:
+  - How do I install cert-manager?
+  - How do I obtain TLS certificates?
+last_review_date: 2022-07-19
+---
+
+Exposing HTTP services using an Ingress is a pretty straightforward task in Kubernetes. Using plain HTTP is, however, discouraged in favor of using HTTPS. `cert-manager` provides a simple, declarative way to automatically obtain TLS certificates for the endpoints we expose.
+
+Our App Platform provides a `cert-manager-app` that users can install. In order to configure it for TLS certificate management, we should install a `ClusterIssuer` that solves DNS challenges. Note that this also works for internal Load Balancers.
```
ubergesundheit, line 16:
```suggestion
Our App Platform provides a [`cert-manager-app`](https://github.com/giantswarm/cert-manager-app) that users can install. In order to configure it for TLS certificate management, we should install a `ClusterIssuer` that solves DNS challenges. Note that this also works for internal Load Balancers.
```

src/content/advanced/tls-certificates/index.md
```diff
@@ -0,0 +1,104 @@
...
+  dnsNames:
+  - "*.cluster.k8s.installation.region.provider.gigantic.io"
+  issuerRef:
+    group: cert-manager.io
+    kind: ClusterIssuer
+    name: letsencrypt-giantswarm
+  secretName: wildcard-tls
+```
+
+this certificate can then be used by setting the following `spec.tls` field in an Ingress:
```
ubergesundheit, line 96: I think this shouldn't be separate.

Usually setting `cert-manager.io/cluster-issuer: nameOfClusterIssuer` annotation on `Ingress` resources makes cert-manager create a `Certificate`

ubergesundheit: mh oh ok, I see this is about the wildcard certificate. Unsure what to do here. Maybe thats okay then to leave it like this

src/content/advanced/ingress/multi-nginx-ic/index.md
```diff
@@ -113,6 +113,20 @@ For NGINX IC running on on-prem (KVM) workload clusters there's no out-of-the-bo
 
 More information on this topic can be found in document [Services of type LoadBalancer]({{< relref "/content/advanced/ingress/service-type-loadbalancer/index.md" >}}).
 
+On AWS and Azure, it is also possible to only install a single Nginx Ingress Controller App and to delegate both internal and external traffic to it. Here is a minimal working example on how to achieve this goal.
```
Gacko, line 116: Theoretically this is also possible on-prem but just results in two different `NodePort` services. This way customers can configure a public and an internal load balancer in front of the workload cluster on their own.

src/content/advanced/ingress/multi-nginx-ic/index.md
```diff
@@ -113,6 +113,20 @@ For NGINX IC running on on-prem (KVM) workload clusters there's no out-of-the-bo
...
+controller:
+  service:
+    public: true  # default value
+    subdomain: "ingress"  # default value
+    internal: 
+      enabled: true  # default is `false`
+      subdomain: "ingress-internal"  # default value
+```
+
+In other words, it is sufficient to set `controller.service.internal.enabled` to `true` to create to services of type `LoadBalancer`: one for public traffic and one for private one.
```
Gacko, line 128: Same here: The specific service type (either `LoadBalancer` or `NodePort`) doesn't matter. The only difference is the way the platform implements the service. In cloud environments you probably go for `LoadBalancer` and let the cloud platform allocate an IP and implement the load balancer, in on-prem environments one would need to create a load balancer pointing to the allocated node ports.

I think even some on-prem platforms already support `LoadBalancer` services, especially when using CAPI. So we could or should just omit the differentiation between `LoadBalancer` and `NodePort` somehow.

src/content/advanced/tls-certificates/index.md
```diff
@@ -0,0 +1,104 @@
...
+  - https://github.com/orgs/giantswarm/teams/team-cabbage
+user_questions:
+  - How do I install cert-manager?
+  - How do I obtain TLS certificates?
+last_review_date: 2022-07-19
+---
+
+Exposing HTTP services using an Ingress is a pretty straightforward task in Kubernetes. Using plain HTTP is, however, discouraged in favor of using HTTPS. `cert-manager` provides a simple, declarative way to automatically obtain TLS certificates for the endpoints we expose.
+
+Our App Platform provides a [`cert-manager-app`](https://github.com/giantswarm/cert-manager-app) that users can install. In order to configure it for TLS certificate management, we should install a `ClusterIssuer` that solves DNS challenges. Note that this also works for internal Load Balancers.
```
Gacko, line 16: The `ClusterIssuer` and `Issuer` CRD comes with `cert-manager`. `cert-manager` creates a `ClusterIsser` using the HTTP challenge by default. This works fine in most cases.

Disabling the creation of this default issuer and creating an own one is probably only require for pure internal environments or for issuing wildcard certificates. So normally you don't need to disable the creation of the default issuer.

Maybe you could point this out here. :)

src/content/advanced/tls-certificates/index.md
```diff
@@ -0,0 +1,104 @@
...
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
+    email: test@customer.com
+    privateKeySecretRef:
+      name: letsencrypt-giantswarm
+    solvers:
+    - dns01:
+        azureDNS:
+          # find the right values in `azure.yaml` in the `external-dns` deployment
```
Gacko, line 42: I think this comment is a bit misleading. Most customers will know both their subscription ID and their cluster ID. So we should not point them to `kubectl exec` into `kube-system` pods.

src/content/advanced/tls-certificates/index.md
Gacko: You probably don't want to issue a wildcard certificate from an ingress because both the `spec.tls.hosts` and `spec.rules.host` must match. So creating a wildcard certificate from you ingress requires this ingress to also match all hosts. But in some cases you just want one wildcard certificate for your (sub) domain and multiple single host name ingresses sharing the same wildcard certificate.

This way `cert-manager` doesn't even need to care about ingress resources and the `cert-manager` annotations can be omitted.

Gacko: Could you also point out that the annotation should be omitted when the certificate has been created "manually"?

Gacko (changes requested):

src/content/advanced/tls-certificates/index.md
LolloneS: Which annotation? `kubernetes.io/tls-acme: "true"`?

src/content/advanced/ingress/multi-nginx-ic/index.md
LolloneS: Done

Gacko (approved):

PR #1507, opened 2022-07-22
Update IRSA doc

T-Kukawka:


webwurst (approved):
very nice to see this so quickly :)

for my taste a bit too much bold, maybe enough to highlight `only right before` and `this is important`, but no hard opinion.

T-Kukawka:
makes sense @webwurst 😅

webwurst (approved):

njuettner (approved):
Thank you

PR #1556, opened 2022-09-14
Kubectl gs deprecated flags update

vvondruska:
### What does this PR do?

- Updated docs of kubectl-gs commands with deprecated flags - added notes about deprecation replacement flags. The affected commands are:
  - `kubectl gs template app`
  - `kubectl gs template catalog`
  - `kubectl gs gitops add app`
- Documented global and root-level flags available in kubectl-gs

Kubectl-gs recently introduced global flags taken over from kubectl. Some of these global flags shadowed command-specific flags with the same names. The shadowed local flags were renamed with the old names still kept around and for backward compatibility and marked as deprecated.

This PR updates the docs to reflect all the changes in kubectl-gs described above.

### DIFF
--- a/src/content/ui-api/kubectl-gs/_index.md
+++ b/src/content/ui-api/kubectl-gs/_index.md
@@ -11,7 +11,7 @@ menu:
   main:
     identifier: uiapi-kubectlgs
     parent: ui-api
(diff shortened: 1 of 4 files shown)


marians (approved):
Great work!

PR #1565, opened 2022-09-19
Add margin to bottom of tab content

marians:
### What does this PR do?

Adds margin to the bottom of tabbed content, to prevent tab content directly "touching" the following content.

### What does it look like?

Before

image

After there are 20 additional pixels of white space.


gusevda (approved):
LGTM

PR #1569, opened 2022-09-22
add documentation for cn-prefix flag in kgs login

anvddriesch:
### What does this PR do?

### What does it look like?


marians (approved):
Just looked at the gsctl docs and found this:

image

We probably have character set restrictions, right?

anvddriesch:
There is no validation in place for this flag right now. Actually AFAIK x509 does not have restrictions in terms of what can be used for common names. However, in practice it might make other components unhappy if we use something random so we might want to add validation. :smile:

PR #1583, opened 2022-10-03
Fix broken links

pipo02mix:
### What does this PR do?

### What does it look like?


marians:
Very nice of you!

I think we should not modify anything in `/changes`, as it's generated and I don't think there is a need to keep links in historic changelogs up-to-date.

pipo02mix:
I have changed those also in the origin, so if is regenerated the values will stay good. My goal is step by step fix all broken links at some point

ubergesundheit (approved):

PR #1587, opened 2022-10-12
Remove CAPZ instructions for `kubectl gs template nodepool` reference

marians:
### What does this PR do?

Removes CAPZ instructions from the `kubectl gs template nodepool` command reference, because

- no customer is using it with CAPZ so far
- we don't want to encourage using this, as CAPZ support will be removed soon
- None of the CAPI providers will be supported by `template nodepool`

### What does it look like?

New provider info:

image

See https://github.com/giantswarm/roadmap/issues/1188

Yes


marians:
Still waiting for review

vvondruska (approved):

PR #1598, opened 2022-10-26
Simplify permissions and add ExternalDNS notice

paurosello:
### What does this PR do?

Improve IRSA docs

Customers did not know about the IRSA implications on ExternalDNS and also simplifying role setup

### DIFF
--- a/src/content/advanced/iam-roles-for-service-accounts/index.md
+++ b/src/content/advanced/iam-roles-for-service-accounts/index.md
@@ -11,7 +11,7 @@ user_questions:
  -  How can I migrate from KIAM to IAM roles for service accounts?
 owner:
   - https://github.com/orgs/giantswarm/teams/team-phoenix
-last_review_date: 2022-08-17
+last_review_date: 2022-10-26
 ---
 
 {{< platform_support_table aws="alpha=v17.2.0" aws="ga=v17.4.0">}}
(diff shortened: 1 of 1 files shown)


src/content/advanced/iam-roles-for-service-accounts/index.md
```diff
@@ -39,23 +39,23 @@ Please first ensure having the following permissions added on the [`GiantSwarmAW
...
       "Effect": "Allow",
       "Action": [
           "s3:PutObjectAcl"
       ],
       "Resource": "arn:aws:s3:::*-g8s-*"
   },
- ...
-```
-
-With AWS release v18.0.0 we additionally need permissions for managing Cloudfront (**only non-China regions**). By upgrading to this AWS release the S3 bucket which contains the OIDC configuration and public keys will be protected and direct traffic will be blocked. We only allow traffic through Cloudfront.
```
T-Kukawka, line 53: why did we remove this part?

whites11 (approved):
lgtmiiwfy

src/content/advanced/iam-roles-for-service-accounts/index.md
paurosello: I think its confusing to have two versions, now it would be 3 versions of the role...

As we just add more things to the role customers should just make sure all of them are in place, or you think its better to keep sepparate?

T-Kukawka: Makes sense👍

T-Kukawka (approved):

PR #1606, opened 2022-10-31
Add highlights for November 3, 2022

marians:
### What does this PR do?

Add weekly highlights for November 3, 2022.

Please add your contributions as commits or suggestions.


gianfranco-l:
done from honeybadger side

marians:
@mproffitt I think "You now have the ability" assumes that the readers are all admins on the management cluster. Which permission is actually required for log access? Can we phrase that in a more specific way?

mproffitt:
@marians would "As cluster admin, the capability now exists..." be a better alternative?

PR #1622, opened 2022-11-14
General / Overview: Point to AWS workload cluster release page

tuladhar:
### What does this PR do?
- Add link to workload cluster release page: https://docs.giantswarm.io/changes/workload-cluster-releases-for-aws

### What does it look like?

Link to workload cluster release page to get more details.

Review.

Yes

### DIFF
--- a/src/content/general/overview/index.md
+++ b/src/content/general/overview/index.md
@@ -43,7 +43,7 @@ There are some things not included in the cluster as managed by us:
 
 ## High availability and resilience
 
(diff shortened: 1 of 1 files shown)


marians:
I don't think that the added link really adds value. If an explanation of the release concept is needed here, I suggest to point to https://docs.giantswarm.io/general/releases/ .

Also please use the [relref](https://gohugo.io/functions/relref/) shortcode syntax for links within the same site.

PR #1639, opened 2022-11-28
Add proxy protocol section in Ingress advanced config

mcharriere:
### What does this PR do?

Add information on how to enable the proxy protocol in `CAPA` and how to disable it in `AWS`

### What does it look like?

image

There is no information on how to set proxy-protocol in `CAPA`, and the current method differs from `AWS`.


ubergesundheit (approved):
Looks good!

PR #1648, opened 2022-12-07
split cluster mgmt into capi and vintage

puja108:
splits cluster mgmt into capi and vintage

also removes overview aka "running Kubernetes on GS" which was not only outdated but also had not much value anymore


pipo02mix:
> Question is if we should move upgrades and releases somewhere else? they still make sense, but need updating along with changes we are still making for CAPI. WDYT?

So the current `Upgrades` page is tied to the CLuster upgrade itself and the current implementation of the release. If we are moving towards a platform where components can be updated at any point in time we should move it (But I will not do it now). There are good recommendations in relation for being more resistant to upgrades (or outages) and those can be moved to `getting-started` in a special document on how to configure your workloads to be more resilent.

carillan81:
Is there a plan to add more sections in `platform-overview/cluster-management/cluster-api/`? 
If no, I would remove the extra step and show directly the platform architecture instead of requiring one more click to go to `platform-overview/cluster-management/cluster-api/architecture/`
Otherwise, I'm fine with the changes

puja108:
> Is there a plan to add more sections in platform-overview/cluster-management/cluster-api/?
If no, I would remove the extra step and show directly the platform architecture instead of requiring one more click to go to platform-overview/cluster-management/cluster-api/architecture/

you're right, I'll change that for now

puja108:
@pipo02mix I agree. I'd see that for a future task, however. For now I'd just move things around a bit so they make a bit more sense, also cause I feel we still have no good picture of howit will work in the future.

pipo02mix (approved):
Quick review, seems ok to me. There is a missing review date to update, rest good

PR #1655, opened 2022-12-09
Management API and web interface overview

teemow:
This moves the management API overview and the web interface documentation to the platform overview section.

Also UIs & APIs have been renamed to "Use the API".


teemow:
All checks have been fixed.

@pipo02mix I left Organizations on the top level for now as a concept. The reason was that 

a) I don't know how you can add sub pages to a page with content properly. They always appear at the end then.
b) Organizations are not that unimportant so I kept it prominent.

pipo02mix (approved):
LGTM in general :)

PR #1656, opened 2022-12-12
Add Support Section 

pipo02mix:
### What does this PR do?

Creates a new section for support and collect some sparsed documents to related to this topic

### What does it look like?

Towards [Public Docs restructurization](https://github.com/giantswarm/giantswarm/issues/20188)

Please check the content of the support overview page, the rest is the same content but reorganized


src/content/getting-started/cluster-size-autoscaling/index.md
```diff
@@ -85,7 +85,7 @@ In workload clusters without autoscaling support, the number of Ingress Controll
 ## See also
 
 - [Cluster Autoscaler advanced configuration]({{< relref "/advanced/cluster-autoscaler" >}})
-- [Recommendations and Best Practices regarding cluster size]({{< relref "/getting-started/best-practices#cluster-sizing" >}})
+- [Recommendations and Best Practices regarding cluster size]({{< relref "/support/best-practices#cluster-sizing" >}})
```
teemow, line 88: Best practices shouldn't be under support imo. They are better within the hands-on guides. "getting-started" sounds good to me.

src/content/support/overview/index.md
```diff
@@ -0,0 +1,77 @@
...
+  - /basics/giant-swarm-support/
+owner:
+  - https://github.com/orgs/giantswarm/teams/team-horizon
+---
+
+In Giant Swarm we have developed a custom support model. In order to offer a great support we define different layers of assistance.
+
+## Direct Support via Slack
+
+Our first level support relies on close interactions via Slack, to ensure bi-directional feedback as quickly as possible. This interaction channel is also used to answer a variety of questions. The conversations are not limited to the platform itself, but can go broader and include the anything cloud-native.
```
teemow, line 25:
```suggestion
Our first level support relies on close interactions via Slack, to ensure bi-directional feedback as quickly as possible. This interaction channel is also used to answer a variety of questions. The conversations are not limited to the platform itself, but can go broader and include anything cloud-native.
```

src/content/support/overview/index.md
```diff
@@ -0,0 +1,77 @@
...
+
+Our first level support shifts run from 08:00 to 18:00 (CET) on Monday-Friday. While production support goes without saying, we  often answer questions well outside those times. We rotate support shifts across our teams, which is why we focus on support channels with clear internal handovers.
```
teemow, line 27: > While production support goes without saying, we  often answer questions well outside those times

What do you mean by "While production support goes without saying"? That we do oncall? And should we really promise that we answer outside the times? Or do you rather want to say that we are in a close relationship with the customers and because of that know what is important when and also support our customers outside of those hours. 

Do we have business hours that we do for US customers? Shall we add those times here too?

src/content/getting-started/cluster-size-autoscaling/index.md
pipo02mix: hmm yeah I don't have a strong feeling, I moving after a discussion with Carlos. I see there we could have content which is not only for starters, but I see it in both sections

src/content/support/overview/index.md
pipo02mix: > While production support goes without saying

This was on the support page already, written by Oli. Let me rewrite it but it wants to emphasize production is 24/7 meanwhile first level is best effort after business hours

puja108 (approved):

PR #1661, opened 2022-12-15
Use url field instead of uri if present

marians:
### What does this PR do?

Use absolute URL for search result item link, if available.

https://github.com/giantswarm/giantswarm/issues/25054


gusevda (approved):
Looks good

PR #1666, opened 2022-12-20
Added connector-id flag to kubectl-gs login

vvondruska:
### What does this PR do?

Documented a new `connector-id` flag supported by kubectl-gs login command and adjusted the documentatoin of the authentication process.

### What does it look like?

There is a new entry in the list of kubectl-gs login flags describing the new flag

Kubectl-gs login command was extended to support multiple identity providers. During the login process it prompts the user to select an identity provider.

A new flag was also added to skip the prompt by providing an identifier of a specific identity provider.

This PR ensures that the above changes are reflected in the documentation

Reviewers can check whether the new entry in the kubectl-gs login documentation makes sense to them

- [x] (Please bump the last_review_date in case this qualifies as a review for an entire page. Provide user_questions which should be answered in the page. Provide a meaningful description.)


src/content/use-the-api/management-api/authentication/user/index.md
```diff
@@ -44,7 +44,15 @@ More information can be found in the [`kubectl gs login`]({{< relref "/use-the-a
...
+In order to complete the authentication flow, `kubectl gs` will open your default web browser. The page in the browser will prompt you to select an identity provider you want to use for the authentication. If you are authenticated with the identity provider you select (determined e.g. via a cookie), you will see a confirmation page and can close the browser window again. However, if you weren't authenticated with your identity provider yet, you'll have to go through the authentication process you are used to. When this is done, a confirmation page will be shown.
+
+You can also execute the `login` command with the `--connector-id` flag, and specify an identifier of your preferred identity provider for the authentication:
+
+```nohighlight
+kubectl gs login URL --connector-id INDENTITY_PROVIDER_ID
+```
+
+When executed with this flag the `login` command does not display the prompt to select an identity provider. Instead, it uses the provider with the specified identifier. In case the provider does not exist, the authentication ends with an error.
```
marians, line 55: I think we should be very careful here with terminology, to keep identity provider and Dex connector apart. With the `--connector-id` flag the user specifies a connector (by its ID), which directly maps to an identity provider.

Another question: As a user, how can I know my connector name, as long as we don't display it in the selection page? I'd add info that there should always be a `customer` connector. And your cluster admin or Giant Swarm support should be able to give out IDs of other connectors.

```suggestion
You can also execute the `login` command with the `--connector-id` flag, and specify an identifier of the desired connector (related to your preferred identity provider) for the authentication:

```nohighlight
kubectl gs login URL --connector-id CONNECTOR_ID
```

When executed with this flag the `login` command does not display the prompt to select an identity provider. Instead, it uses the provider with the specified identifier. In case the provider does not exist, the authentication ends with an error.

**Note:** Connector identifiers are currently not displayed. If you have only one connector configured, it should have the identifier `customer`. To find out identifiers of additional connectors, please contact Giant Swarm support.
```

File: src/content/use-the-api/kubectl-gs/login.md [hunk 2]
```diff
@@ -16,7 +16,7 @@ user_questions:
   - How can I log in with kubectl for the Management API?
   - How can I create a workload cluster client certificate?
   - How do I specify the time to live for a workload cluster client certificate?
-last_review_date: 2022-05-20
+last_review_date: 2022-12-20
```
marians, line 19:
```suggestion
last_review_date: 2023-01-10
```

marians:
Here is an update of the schema image: https://docs.google.com/drawings/d/1UD1Pw9UxzCEO7WMtp57MOizkVqU6Oq5IiUdTbhU0VWo/edit

You can download the PNG from there.

vvondruska:
Thanks. Schema image updated.

anvddriesch (approved):

PR #1688, opened 2023-01-24
Adding a link on how to create the secret with Flux

pipo02mix:
Adding a link on how to create the secret with Flux

### What does this PR do?

### What does it look like?


carillan81 (approved):
nice add. This was pending as the instructions were in the intranet but this link is exactly what we need. Thanks!

PR #1691, opened 2023-01-26
contribution kgs add base for weekly highlights

gianfranco-l:
Contribution about gitops add base command for weekly highlights

### DIFF
--- a/src/content/changes/highlights/2023-01-26.md
+++ b/src/content/changes/highlights/2023-01-26.md
@@ -16,7 +16,7 @@ owner:
 
 ## User Interfaces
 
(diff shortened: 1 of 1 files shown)


uvegla:
It looks good to me, but I think the docs needs to be updated because of: https://github.com/giantswarm/kubectl-gs/blob/main/CHANGELOG.md#2310---2023-01-18 the folder structure was slightly changed. (Also might want to link this too.)

gianfranco-l:
@carillan81 every week we create a list of highlights related to updates to docs, user interfaces and managed apps to our customers. This communication is sent out every thursday, in the early afteroon (Around 2 pm CET). About gitops add base command, do you think we can communicate this to customers already via this communication (that goes out to all customers Slack channels) or we need to hold on a bit until the change suggested by @uvegla is done and maybe postpone it for next week? Pls let me know :)

carillan81:
I am working in the environments tutorial now, so we might wait and announce them together.

gianfranco-l:
ok then, I'll remove the reference for gitops add base for this week, @carillan81 let us know once this piece of doc is ready to be announced

PR #1694, opened 2023-02-01
Deprecate KVM crds

cornelius-keller:
### What does this PR do?
- deprecate KVM crds

### What does it look like?

Suppor for KVM clusters has reached end of live and will be discontinued in the future.

- check if my PR is technically correct
- check wording for deprecation notice

### DIFF
--- a/scripts/update-crd-reference/config.yaml
+++ b/scripts/update-crd-reference/config.yaml
@@ -363,7 +363,7 @@ source_repositories:
         topics:
           - workloadcluster
         deprecation:
-          info: Support for KVM clusters  has reached end of life  and this CRD will be removed in future versions.
+          info: Support for KVM clusters has reached end of life and this CRD will be removed in future versions.
       machinedeployments.cluster.x-k8s.io:
         owner:
           - https://github.com/orgs/giantswarm/teams/team-phoenix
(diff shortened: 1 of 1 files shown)


marians (approved):

scripts/update-crd-reference/config.yaml
```diff
@@ -353,13 +353,17 @@ source_repositories:
           - kvm
         topics:
           - workloadcluster
+        deprecation:
+          info: Support for KVM clusters  has reached end of life  and this CRD will be removed in future versions.
```
marians, line 357:
```suggestion
          info: Support for KVM clusters has reached end of life and this CRD will be removed in future versions.
```

scripts/update-crd-reference/config.yaml
```diff
@@ -353,13 +353,17 @@ source_repositories:
...
       kvmconfigs.provider.giantswarm.io:
         owner:
           - https://github.com/orgs/giantswarm/teams/team-rocket
         provider:
           - kvm
         topics:
           - workloadcluster
+        deprecation:
+          info: Support for KVM clusters  has reached end of life  and this CRD will be removed in future versions.
```
marians, line 366:
```suggestion
          info: Support for KVM clusters has reached end of life and this CRD will be removed in future versions.
```

marians:
Still approved :)

PR #1709, opened 2023-02-22
Add Prometheus Agent documentation

QuentinBisson:
### What does this PR do?

Add first documentation of the prometheus agent for customers

### What does it look like?

We had to deploy a prometheus agent in clusters to be able to monitor kubernetes core components. This pages describe what it is, why we added it and the architecture.

An overall help in making this better

Done


QuantumEnigmaa (commented):
In your diagram you wrote `workload cluster prometheus` for the one in the MC. But isn't the point of using agents to get rid of the prometheus pods for each WC and only having a central one in the MC on which each agent will access using Remote Write ?

QuentinBisson:
So no the plan is to go with mimir in the end and not a single Prometheus because it would use too much memory and CPU.

Also, this doc reflects the current status as I cannot really write either Mimir or LTS as our prometheus is currently not LTS :( Maybe @TheoBrigitte or @hervenicol have opinions :)

hervenicol:
Totally agree with @QuentinBisson:
* we should show current status
* long term goal is a "scalable prometheus-like" like Mimir, but prometheus is still here for a while

hervenicol:
Concerning this documentation, I think it would be useful to have this info:
* data from agent to prometheus goes through an ingress
* metrics are in prometheus-remote-writes and nginx dashboard
* maybe some links to related opsrecipes (like https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/prometheus-agent-remote-write-failed/)
* we can see web-up with the list of targets on the agent with a port-forward
* agent scrapes servicemonitors that have the label `giantswarm.io/team`
* prometheus keeps scrapes through the "old-style" scraping, ie `giantswarm.io/monitoring` label.

QuentinBisson:
@hervenicol sure but this is external documentation to customers. Do we want to go into that much details especially regarding ops-recipe? I will add information about the dashboards and servicemonitors though

hervenicol:
Maybe not all of it in the public doc, but I think all of this should be documented.
As the opsrecipes are private, sure, not links to opsrecipes.
But for all the rest, would you rather split the info in 2 docs, private / public, or have it all on the public one? Is there anything "secret" in it?

QuentinBisson:
There are no secrets no but I'm afraid a lot of it would be confusing because the basic of our monitoring setup is not documented :D

QuentinBisson:
@hervenicol can you take another look? I think I added all your points :)

QuentinBisson:
We can for sure improve it :)

QuentinBisson:
Also thanks a lot for taking the time to review :)

src/content/getting-started/observability/monitoring/prometheus/agent/index.md
```diff
@@ -0,0 +1,86 @@
...
+
+As the Prometheus Agent is running in your workload cluster, you can gain access to its UI to inspect the list of currently configured targets it is scraping using the following command `kubectl port-forward -n kube-system prometheus-prometheus-agent-0 9090` then access the url `http://localhost:9090/targets` on your favorite browser.
+
+You should see the following:
+
+![Screenshot of Prometheus Agent User Interface](prometheus-agent-ui.png)
+
+### Access to the metrics
+
+The metrics of the Prometheus Agent can be found on the [grafana instance]({{< relref "/getting-started/observability/grafana/access" >}}) running on your management cluster under the `Prometheus / Remote Write` and `NGINX Ingress controller` dashboards
```
hervenicol, line 86: Just checked, "prometheus/remote write" dashboard is private :cry:

hervenicol (approved):
I love it! Plenty of useful info :grinning: 
As for any doc, there's still room for improvement, but it's already good enough and we can improve over time.

src/content/getting-started/observability/monitoring/prometheus/agent/index.md
QuentinBisson: I have a pr on dashboards to change this but I need to test it 😂

pipo02mix:
ye4ah good work, sorry but I arrived late

PR #1710, opened 2023-02-23
add promtail audit log config link

TheoBrigitte:
### What does this PR do?

Add the missing link for Promtail audit log configuration example.

Is this link fine or should we instead link directly to the concerned file at https://github.com/giantswarm/promtail-app/blob/master/examples/values-kubernetes-audit-logs.yaml ?


QuentinBisson:
I think it's fine ;)

QuentinBisson (approved):

PR #1720, opened 2023-03-06
Fix issue title

pipo02mix:
### What does this PR do?

Fix a bug with Issue Generation. The title was empty before

### What does it look like?


.github/workflows/generate-review-issues.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
           echo "ISSUES=$(cat files.txt | python3 scripts/validate-front-matter/script.py --validation last-reviewed --output json)" >> $GITHUB_ENV
       
       - name: Generate issues of out of date docs
-        uses: giantswarm/open-issue@0.1.0
+        uses: giantswarm/open-issue@0.3.0
```
pipo02mix, line 25: newer version fix GH API rate limiting

PR #1724, opened 2023-03-13
Updated 'OIDC auth for workload clusters' docs page

vvondruska:
### What does this PR do?

Review and update of the 'OIDC auth for workload clusters' docs page, introducing the following changes:

- example of OIDC configuration in a CAPI cluster
- adjustment in the part about deploying the Dex app to use a Secret instead of a ConfigMap to provide the configuration values
- additional section about custom certificates in Dex
- additional section about running Dex in a private workload cluster

### What does it look like?

Just textual changes

Dex app has a few recent changes, which need to be reflected in the documentation.

Check if the documentation makes sense

Last review date updated
The new content expands answers to the existing user questions (mostly "How can I configure OIDC in my cluster?")

### DIFF
--- a/src/content/advanced/configure-dex-in-your-cluster/index.md
+++ b/src/content/advanced/configure-dex-in-your-cluster/index.md
@@ -9,7 +9,7 @@ menu:
 user_questions:
   - How can I configure OIDC in my cluster?
   - How can I add a new OIDC connector?
-last_review_date: 2022-03-04
+last_review_date: 2023-03-13
 owner:
   - https://github.com/orgs/giantswarm/teams/team-rainbow
 ---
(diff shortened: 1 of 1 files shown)


src/content/advanced/configure-dex-in-your-cluster/index.md
```diff
@@ -261,29 +260,63 @@ metadata:
...
+  tls:
+    letsencrypt: false
+    caPemB64: ...
+    crtPemB64: ...
+    keyPemB64: ...
+```
+
+### Running Dex in a private workload cluster
+
+When deploying Dex to a private workload cluster, it is required to specify a proxy configuration in order to ensure that Dex has access to the outside network. Proxy setup can be provided in the `proxy` section of the `Secret` or `ConfigMap` with configuration values for the Dex app. It is required to set the proxy as `enabled`, specify the URL of the HTTPS proxy in the `https_proxy` property and exclude Kubernetes API from running through the proxy by adding its IP address or network to the `no_proxy` property.
```
anvddriesch, line 310: Consider making new lines after sentences. When someone makes changes later, the PRs are tidier that way.

anvddriesch (approved):

PR #1727, opened 2023-03-13
Add owner to issue generator script

pipo02mix:
### What does this PR do?

I found a bug in the last issue generation for stale docs and the label was missing. I need to convert the metadata field from `team-xxxx` to `team/xxxx`

### What does it look like?


stone-z (approved):
WFMIIWFY

PR #1776, opened 2023-04-11
Document GCP internal load balancer annotation

ubergesundheit:
This PR adds the Service annotaton to request an internal load balancer on GCP


Gacko (approved):
Quotes not needed here, but fine. LGTM! :)

PR #1810, opened 2023-05-17
fix capz example to make use of cluster-values

bavarianbidi:
### What does this PR do?

### What does it look like?

with CAPZ based clusters with the new `kube-proxy`-less mode, the correct `baseDomain` is required. This value is stored in the `cluster-values` configmap and needs to be referenced. The change is already implemented in `kubectl-gs`


primeroz (approved):
Should we also add the `managed-by` label ?

AndiDog:
I suggest we close this PR in favor of https://github.com/giantswarm/docs/pull/1836, since that one deletes the article altogether in order to merge all provider instructions into one article.

PR #1826, opened 2023-06-02
Provoke front matter validation error

marians:
This PR is only a demonstration of front matter validation as a GitHub action.


pipo02mix:
Do we still need this?

PR #1831, opened 2023-06-07
Reorder "Getting started" articles, remove differing link titles, rename "Getting > {Best practices => Cluster sizing, multi-tenancy and autoscaling}"

AndiDog:
### What does this PR do?

Apart from the title, I'm changing `make dev` to run hugo locally, since we found in chat that the current Docker-based workflow doesn't work. Running it locally is faster and easily works with file watching.

### What does it look like?

This matches the [drafted order](https://docs.google.com/document/d/1KpYOnswKOjWIohba6l7u3HUlYJo2s0ZTI5uKNsLjNWE/edit#) of "Getting started", renames one article, but otherwise doesn't introduce major changes yet. I've used multiples of 10 so we can easily add more articles in between if needed. Some numbers are missing as placeholder for the articles that we want to add soon. Here's the current order with my changes:

```text
λ grep -r weight: src/content/getting-started/*/*index.md | awk '{print $NF " - " $1}' | sed -E 's|src/content/getting-started/([^/]+).*|\1|' | sort -n
20 - cloud-provider-accounts
40 - kubectl-gs
50 - sizing-multi-tenancy
60 - app-platform
70 - ingress-controller
80 - exposing-workloads
90 - persistent-volumes
100 - observability
110 - rbac-and-psp
120 - operations
130 - ca-certificate
140 - network-policies
150 - capa-wlcluster-creation
160 - capz-cluster-creation
```

https://github.com/giantswarm/roadmap/issues/2393

No, since content doesn't change.


src/content/getting-started/app-platform/_index.md
```diff
@@ -1,6 +1,5 @@
 ---
-linkTitle: Deploy applications using App platform
-title: Using App platform to deploy applications
+title: Deploy applications using the App platform
```
pipo02mix, line 2: but if we remove `linkTitle` front matter fails. Also in some occasion we want to have different menu title than page title, don't you think?

Maybe we can do that when there is no linktitle use title but it will need a change in the script

pipo02mix (commented):
LGTM but the linktitle needs to be fixed (or change the script or add it again)

src/content/getting-started/app-platform/_index.md
AndiDog: What exactly fails? Hugo source [code](https://github.com/gohugoio/hugo/blob/73779707a2d5fefd4915f8ee2021f265d841dea2/hugolib/page__meta.go#L182) and docs say that it defaults to `.Title` without requiring an explicit `{{ or .LinkTitle .Title }}` in any template.

I don't see anything wrong in the rendered output at the moment. Except for this which seemed to be there before my changes:

Screenshot 2023-06-07 at 14 22 39

pipo02mix:
Fails the checker 

Screenshot 2023-06-07 at 15 34 54

src/content/getting-started/app-platform/_index.md
pipo02mix: yeah the overview section should not exists, not sure what happened. I will do a PR

pipo02mix (approved):

PR #1834, opened 2023-06-08
Update sysctl examples

stone-z:
Just updating the example Pods for the sysctl policy. Those fields require values as well as names, so the example was incorrect.


fhielpos (approved):
Works for me

PR #1855, opened 2023-07-03
Highlight selected item in menu

AndiDog:
### What does this PR do?

This fixes a main annoyance with the menu – the currently selected article _does_ have a somewhat different color, but it's barely distinguishable and hence the reader may not see which page is active.

### What does it look like?

Switch to a more link-like color and making it slightly bold (not fully bold because the main articles such as "Platform overview" are already bold):

Screenshot

No


pipo02mix (approved):
nice

PR #1856, opened 2023-07-04
Move implementation titles into site parameters so they can be consistently used as tab titles

AndiDog:
### What does this PR do?

I wanted to have consistent tab titles. Since you cannot nest shortcodes like this

```
{{< tab id="cluster-vintage-azure" title="{% cluster_mgmt_impl_vintage_azure %}">}}
```

there is now a translation layer using site parameters. So we can change the titles in exactly one spot.

### What does it look like?

No changes to rendered text

https://github.com/giantswarm/roadmap/issues/2393

No

### DIFF
--- a/src/config.yaml
+++ b/src/config.yaml
@@ -53,3 +53,19 @@ taxonomies:
 
 permalinks:
   changes_categories: /changes/:slug
+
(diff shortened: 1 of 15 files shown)


pipo02mix (approved):
LGTM. This even better than shortcodes ❤️

PR #1859, opened 2023-07-06
Update Kyverno PolicyExceptions docs

fhielpos:
@stone-z can you use some of your English wizardy to make this prettier? I don't want to sound redundant since the previous points talks about the same


fhielpos:
Actually, I don't know if it doesn't make more sense to use a `Deployment` or `CronJob` as an example instead. WDYT?

stone-z:
Yeah since there's been confusion it might be helpful to demonstrate the whole Deployment+ReplicaSet+Pod flow

src/content/advanced/security-policy-enforcement/index.md
```diff
@@ -1093,6 +1093,7 @@ This example allows a DaemonSet (and the Pods it creates) named `my-workload` in
 Noteworthy pieces of this example:
 
 - Kyverno policy rules are usually written at the Pod level. For convenience, Kyverno automatically generates equivalent rules for Pod controllers like Deployments and DaemonSets. Such rules are prefaced with the value `autogen-` and added to the policy automatically (two such rules are visible in the example). When writing a `PolicyException`, any applicable `autogen` rules must also be listed if a workload should be exempt from them.
+- When listing resource kinds in a `PolicyException`, every subresource controller must be listed. For example: If a Policy is written at the CronJob level, then the Job and Pod controllers need to be explicitly defined in the `PolicyException`. The same happens with Deployments, where the ReplicaSet and Pod controllers will need to be excluded too.
```
stone-z, line 1096: These could almost just be combined since they're two sides of the same behavior, but minor suggestion with them as-is:
```suggestion
- Kyverno policy rules are usually written at the Pod level. For convenience, Kyverno automatically generates equivalent rules for Pod controllers like Deployments and DaemonSets. Such rules are prefaced with the value `autogen-` and added to the policy automatically (two such rules are visible in the example). When writing a `PolicyException`, any applicable `autogen` rules must also be listed if a workload should be exempt from them.
- Similarly, when listing resource kinds to be matched in a `PolicyException`, every subresource controller must be listed as well. For example: If a Policy is written at the CronJob level (and autogen policies are enabled for it), then the Job and Pod resources created from the CronJob also need to be explicitly matched in the `PolicyException`. The same happens with Deployments, where the ReplicaSet and Pod controllers will need to be excluded as well.
```

fhielpos:
Updated ✅

stone-z (approved):
Beautiful

PR #1879, opened 2023-08-10
Mermaid support

pipo02mix:
### What does this PR do?

It adds mermaid support for render diagrams and fix an existing broken mermaid code in RBAC page.

### What does it look like?


README.md
```diff
@@ -10,6 +10,10 @@ Please review the [Contribution guidelines](CONTRIBUTING.md) for guidelines on s
 
 If you are an employee of Giant Swarm you can get more information per [intranet page](https://intranet.giantswarm.io/docs/product/docs/) with info on style and other pre-requisites.
 
+## Diagrams
+
+We use [mermaid]() for diagrams. You need to annotate the page previously in the frontmatter to load the mermaid code (`mermaid: true`). Then you can use the shortcode `{{< mermaid >}}` to add the diagram code.
```
stone-z, line 15:
```suggestion
We use [mermaid](https://mermaid.js.org/) for diagrams. You need to annotate the page previously in the frontmatter to load the mermaid code (`mermaid: true`). Then you can use the shortcode `{{< mermaid >}}` to add the diagram code.
```

stone-z (commented):
I think `mermaid` also needs to be added to the known front matter keys https://github.com/giantswarm/docs/blob/main/scripts/validate-front-matter/script.py#L198

stone-z (approved):
:heart:

PR #1886, opened 2023-08-25
Fix markdown

webwurst:
Greetings! :)


AndiDog (approved):

AndiDog:
This will anyway be fixed alongside hundreds of other lints. See https://github.com/giantswarm/docs/pull/1888 which fixes the markdownlint pipeline. So I'll close this because it will be part of another PR. Thanks 😉

PR #1888, opened 2023-08-28
Fix Markdown linting and annotate in PRs

AndiDog:
### What does this PR do?

The CircleCI job didn't do anything. And since it's nicer to see lints annotated into the PR diff, I moved linting to a GitHub action. The `make lint` command should now be fully usable by developers and CI.

### What does it look like?

This PR should show lots of errors. _Note_ that I have all the lint fixes ready, but that would go into a huge, separate PR mostly about `_` vs. `*`, `-` vs. `*`, many whitespace changes, and some heading fixes. So for _this one time_, we should ignore the pipeline failure of _this PR_, merge it, and I'll open a new PR to fix all linting findings.

Why errors and not warnings? Well, we piled up lots of lint errors and that shouldn't happen in the future. PRs pipelines should show a clear failure. That way, we get consistent articles, and people will configure markdownlint in their editor (e.g. [VS Code extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) + trim trailing whitespace on save) to see and prevent issues already at time of writing.

### Any background context you can provide?

I didn't like that Markdown files had trailing whitespace 🧠🤯 and kept going 🤣.

PR #1891, opened 2023-08-28
Fix all Markdown linter findings

AndiDog:
### What does this PR do?

Follow-up to #1888

### What does it look like?

All green


pipo02mix (approved):
LGTM

PR #1893, opened 2023-08-29
Remove or mark REST API as deprecated in some places

AndiDog:
### What does this PR do?

Since we're removing the REST API anyway [really soon](https://gigantic.slack.com/archives/C02521EC0/p1693215237926119), I wanted to already mark it deprecated so that customers reading up about Cluster API won't even read those parts or get distracted. This PR makes a small start in that direction.

Date updated only for articles where I had knowledge and time to refresh them entirely.


gawertm (approved):

pipo02mix (approved):
nice

PR #1907, opened 2023-09-13
Remove dex-k8s-authenticator

mogottsch:
### What does this PR do?

This PR removes all references to dex-k8s-authenticator from the docs, as it is deprecated.

Towards https://github.com/giantswarm/roadmap/issues/2783


anvddriesch (approved):

marians (changes requested):
In the intro, there is still this paragraph talking about two methods:

> The recommended method for authentication, as it's the fastest and most convenient one, uses our `kubectl` plug-in and is explained next. Further down we also provide instructions for an [alternative method](#alternative-method) that does not require installing the plugin.

mogottsch:
Good catch @marians !
I've updated the paragraph and removed the reference to the "alternative method".

AndiDog (approved):
Thanks for cleaning that up!

marians (approved):
Thanks!

PR #1921, opened 2023-09-26
Updating CRDs doc

ljakimczuk:
### What does this PR do?

It updates App CRD doc page deprecating one field.

### What does it look like?

Add new `Deprecated` property.

Context: https://github.com/giantswarm/roadmap/issues/2830

Does this way of visualization work for you?


pipo02mix (approved):
cannot see the final view but LGTM

marians:
Actually the CRD docs cannot be changed in this way, as they are generated. I told @ljakimczuk already via Slack.

PR #1929, opened 2023-10-06
Add Policy API and PSP migration docs

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/28243
### What does this PR do?

Adds dedicated pages about the Policy API and the migration from PSPs to PSS.

A second update will include a more detailed sample workflow using the Policy API objects instead of Kyverno PolicyExceptions, but since we are generating Draft exceptions, we don't expect much immediate need to write them by hand.

### What does it look like?

https://github.com/giantswarm/roadmap/issues/2716
https://github.com/giantswarm/roadmap/issues/256

This guide describes the expected migration workflow for customers to use when moving from PSP-secured clusters to PSS-secured clusters using the new Policy API.

Yes!


src/content/advanced/security-policy-enforcement/cluster-admin-guide.md
```diff
@@ -0,0 +1,116 @@
...
+ -  How can I exclude a workload from a Kyverno policy?
+ -  What security policies are enforced in my cluster?
+ -  How do I migrate from PSPs to PSS?
+ -  What is the Policy API?
+owner:
+  - https://github.com/orgs/giantswarm/teams/team-shield
+last_review_date: 2023-10-04
+---
+
+<!-- {{< platform_support_table aws="alpha=v17.2.0" aws="ga=v17.4.0">}} -->
```
AndiDog, line 21: If it's commented out, do we need it? Shouldn't it say AWS v19.x?

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+__Note__: This guide is intended for cluster administrators running Giant Swarm managed Kubernetes clusters. More general information about Pod Security Standards can be found on the [Security policy enforcement][sec-policy-enforcement] page.
+
+## Managing cluster security policies with the Giant Swarm Policy API
+
+The Policy API is an abstraction layer which orchestrates several other types of policy-related resources.
+
+The Giant Swarm platform is built on top of a number of independent tools, projects, and APIs supported by the CNCF and the surrounding Kubernetes ecosystem.
+Among these are a number of resources designed for enforcing policies within a cluster. These include built-in types like Network Policies and RBAC as well as external CRDs like Kyverno Cluster Policies or Cilium Network Policies, among others.
+
+We use and manage tools which we see as bringing value for customers, and which are governed and maintained in a way which aligns with our guiding principles.
```
AndiDog, line 32: Can we shorten this 3x"We" paragraph? Let's make it very explicit and use examples. I guess what we want to say here is that Kyverno and PSS is the current thing in CNCF landscape, but since it may change in the future, customers have the choice to use our high-level "I want app X to have restriction Y" policies over low-level Kyverno policies. We can then describe the benefits of our high-level CRDs as separate paragraph ("without requiring customers to change tools [...]").

What I mean is that this and the next paragraphs may be too verbose for a customer's engineer to read.

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+- provides a way for cluster administrators to declare their intent about which policies to enforce and which resources are exempt from those policies.
+- is intended to manage additional policy types in the future, including networking, vulnerability management, anomaly detection, and others.
+- is _not_ a general purpose policy syntax language. Users cannot define custom policies via the Policy API. It can only be used to configure the policies Giant Swarm actively manages.
+- does _not_ hide the underlying implementations. Users are free to directly use the underlying tools or APIs. The only difference is that _Giant Swarm will not manage, migrate, or adopt any policies or configuration passed directly_.
+- generates native resources for the underlying implementations. These resources continue to function even if the Policy API controllers are removed.
+  
+#### Managed and un-managed policies
+
+The Policy API only orchestrates policies which Giant Swarm actively manages. It does not interfere with customer policies, exceptions, or configuration which are managed externally to the Policy API.
```
AndiDog, line 55: > The Policy API only orchestrates policies which Giant Swarm actively manages.

As a customer, I wouldn't really get the point. Rather than saying "Giant Swarm manages something", can we describe it as "Giant Swarm provides a set of ready-made policies to choose from"? At least that's how I understand this works. And by now, we're already a few minutes into someone reading the article, so let's put an example here.

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+For example, Giant Swarm enforces security policies in every cluster by default, and advises customers to use the Policy API to declare exceptions for any workloads which need them.
+We currently use Kyverno to enforce those policies, and the Policy API generates Kyverno PolicyExceptions based on the exceptions configured through the Policy API.
```
AndiDog, line 58:
```suggestion
We currently use Kyverno to enforce those policies, and the Policy API generates Kyverno `PolicyException`s based on the high-level exceptions configured through the Policy API. If Kyverno needs to be replaced by a better solution later, you do not need to make any changes.
```

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+A cluster administrator might choose to use that managed Kyverno to enforce their own additional policies, to enforce some business-specific logic, for instance.
+They can easily do that by creating a new Kyverno ClusterPolicy, and creating Kyverno PolicyExceptions for any approved exceptions.
```
AndiDog, line 61:
```suggestion
They can easily do that by creating a new Kyverno `ClusterPolicy`, and creating Kyverno `PolicyException`s for any approved exceptions.
```

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+If, hypothetically, in the future, Giant Swarm chooses to stop managing Kyverno as part of our standard platform, we would use our Policy API controllers to move our managed policies and any relevant exceptions into a new implementation which maintains the desired behavior.
```
AndiDog, line 63: See above review comment which suggests the important point – customers don't need to change anything in case of migration. If we apply the above suggestion, this sentence _here_ can go away.

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+The custom ClusterPolicy, and any configured Kyverno PolicyExceptions, would need to be adapted by the cluster administrator, or they would need to then manage Kyverno themselves.
```
AndiDog, line 65: Let's put this sentence together with "A cluster administrator might choose to use that managed Kyverno to enforce their own additional policies" above, and separate this into an "unmanaged" section to not create back-and-forth confusion when reading through the paragraphs.

AndiDog (changes requested):

src/content/advanced/security-policy-enforcement/policy-api.md
stone-z: The information follows the same chronological progression I'm trying to highlight. Open to other feedback here, but it was very intentional to present the current state followed by the future possibilities, since that's what I want to convey

src/content/advanced/security-policy-enforcement/policy-api.md
stone-z: Commented above

src/content/advanced/security-policy-enforcement/policy-api.md
stone-z: Commented above

src/content/advanced/security-policy-enforcement/policy-api.md
stone-z: Well, the next sentence is an example. But I can reword this
```suggestion
Giant Swarm provides a set of ready-made policies for many common cluster management use cases. The Policy API only orchestrates these standard policies, which Giant Swarm actively manages. It does not interfere with customer policies, exceptions, or configuration which are managed externally to the Policy API.
```

src/content/advanced/security-policy-enforcement/policy-api.md
```diff
@@ -0,0 +1,69 @@
...
+We want to make that overhead optional, and be able to fully manage a feature without requiring customers to change tools, update config, or plan for complicated migrations when one of the underlying implementations changes.
+
+For example, when the Pod Security Policy type was deprecated and removed, many workloads had to be re-evaluated and new exceptions created for them, even though neither the workload nor the risk had changed.
+Some new features were gained, but the effort was primarily invested in order to maintain feature parity and avoid security regressions.
+We expect that Kubernetes and third-party tooling will continue to evolve, and that we can help our customers be faster if they are not directly tied to tool-specific interfaces which they don't actually want to manage.
+
+So, we created the Policy API in order to allow Giant Swarm to more seamlessly and transparently move clusters between policy implementations, and to reduce the overall toil of dealing with common security configuration.
+
+### At a glance
```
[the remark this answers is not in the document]
stone-z: Sure, I can move it up

src/content/advanced/security-policy-enforcement/policy-api.md
stone-z: Yeah I also don't like the We x 3, left it for future me in the next update.

This paragraph overall though is supposed to be general about the policy API. The current use cases and policies will be described in the lower sections, this is more to explain the motivation for even having the API and admittedly to put folks at ease about something so GS specific. Basically I'm toeing the line between wanting customers to use everything we provide and telling them to be realistic about what they want to manage

stone-z (commented):
Need to step away for a bit but will move some more things around to address other comments later

fhielpos (approved):

src/content/advanced/security-policy-enforcement/index.md
```diff
@@ -20,20 +20,25 @@ last_review_date: 2023-03-21
 
 <!-- {{< platform_support_table aws="alpha=v17.2.0" aws="ga=v17.4.0">}} -->
 
-__Note__: this content is applicable only to clusters running Kyverno, with the PSS `baseline` and `restricted` policies deployed in `enforce` mode. This will become the default in future Giant Swarm clusters, but can already be enabled on demand. This page will be updated to indicate the platform version where this content becomes the standard.
+__Note__: this content is applicable to Giant Swarm v19.2.0 and above, which include PSS `baseline` and `restricted` policies deployed in `enforce` mode.
 
-To enforce security best practices, several Kyverno policies mapped to the [Kubernetes Pod Security Standards][k8s-pss] are pre-installed in Giant Swarm clusters.
+__Note__: additional information for cluster admins can be found in a separate cluster admin guide.
```
pipo02mix, line 25: What about adding the link?

stone-z: Yeah I originally added it, but markdownlint blows up because the page doesn't exist yet :sweat_smile: I'll add it in the second PR
https://github.com/giantswarm/docs/pull/1929#issuecomment-1750955093

AndiDog:
> @AndiDog I'd like to get this published since the relevant release is likely going out today -- any more immediate blockers from your side or can we address anything else in followups?

I'm fine if the leftovers get addressed in a follow-up.

AndiDog (approved):

PR #1936, opened 2023-10-12
Document private cluster creation for CAPA

AndiDog:
### What does this PR do?

This was missing from https://github.com/giantswarm/roadmap/issues/2393. There may still be other adjustments to make such as tagging subnets for specific uses (`subnet.giantswarm.io/endpoints`, `subnet.giantswarm.io/control-plane`, etc.), but let's at least get the basic command documented for now.

Also, I reconsidered the `kubectl gs template cluster --name` parameter and made it explicit in our examples, since nobody should use random names.

The `--dns-mode` parameter is gone since we removed support for private DNS mode in CAPA clusters.

yes


calvix (approved):
lgtm

PR #1946, opened 2023-10-25
update rbac doc with matchExpressions examples for roleBindingTemplates

anvddriesch:
Towards https://github.com/giantswarm/roadmap/issues/2880

### What does this PR do?

### What does it look like?


scripts/update-crd-reference/config.yaml
```diff
@@ -511,7 +511,7 @@ source_repositories:
   - url: https://github.com/giantswarm/rbac-operator
     organization: giantswarm
     short_name: rbac-operator
-    commit_reference: v0.37.1
+    commit_reference: v0.39.0
```
anvddriesch, line 514: This is not released yet. We can only merge it once it is.

vvondruska (approved):

PR #1956, opened 2023-10-30
PDB caveat

vxav:
### What does this PR do?

### What does it look like?


src/content/platform-overview/cluster-management/cluster-upgrades/index.md
```diff
@@ -214,6 +214,8 @@ If a pod takes a long time for a proper shutdown, configure the `terminationGrac
 
 Configure [PodDisruptionBudgets](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for all your deployments. This tells Kubernetes to keep a minimum amount of Pods running at all times and is respected by the draining/eviction mechanisms during upgrades.
 
+Note that, in the case of deployments where `replicas=1`, setting a `PodDisruptionBudgets` with `maxUnavailable=0` is not advised as it will hold the cluster upgrade since the node won't be able to be fully drained.
```
pipo02mix, line 217:
```suggestion
__Note__: In the case of deployments where `replicas=1`, setting a `PodDisruptionBudgets` with `maxUnavailable=0` is not advised as it will hold the cluster upgrade since the node won't be able to be fully drained. Depending on the provider there will be a different timeout that will make pod to be force stopped and evicted from the node.
```

pipo02mix (approved):

PR #1960, opened 2023-11-03
Add missing line in internal NGINX docs

LolloneS:


mcharriere (commented):
would you add the `*.` also here?:
```
subdomain: ingress-internal
```

LolloneS:
yeah let's do it, I applied the change

mcharriere (approved):

PR #1973, opened 2023-11-15
Skip emitting change Markdown file for empty release body in `docs` repo, make `date` format consistent

AndiDog:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/2962

On the next run, this removes existing files which match the given list of "considered empty" bodies, and doesn't emit new such files.

During manual testing, I found that the `date` field is rendered inconsistently since sometimes, a timezone-aware `datetime` object is parsed and that would add the `+00:00` suffix which we don't really want. Made that consistent.

### DIFF
--- a/scripts/aggregate-changelogs/script.py
+++ b/scripts/aggregate-changelogs/script.py
@@ -216,7 +216,7 @@ def link_commit_hashes(mkdwn, repo_shortname):
     result = re.sub(r'\b([a-f0-9]{40})\b', replace_hash, mkdwn)
     return result
 
-def generate_release_file(repo_shortname, repo_config, release):
+def generate_release_file(repo_shortname, repo_config, release, delete):
     """
     Write a release file with YAML front matter and Markdown body
     """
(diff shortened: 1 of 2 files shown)


scripts/aggregate-changelogs/script.py
```diff
@@ -300,8 +305,9 @@ def generate_release_file(repo_shortname, repo_config, release):
         catfile.write(dump(categories, Dumper=CDumper))
 
     # Write changelog items
-    for repo_short in conf['repositories']:
-        repo_conf = conf['repositories'][repo_short]
+    for repo_short in sorted(conf['repositories']):
+        print(f'Repo {repo_short}')
```
pipo02mix, line 309: do you want to keep this one?

pipo02mix (approved):
LGTM

scripts/aggregate-changelogs/script.py
AndiDog: Yes. It helps with manual testing to see which repo is processed. That's also why I changed to `sorted` for iteration in order to achieve a stable order.

PR #1977, opened 2023-11-16
Remove opencost from docs

njuettner:
We don't support opencost anymore

Issue: https://github.com/giantswarm/roadmap/issues/2785


pipo02mix:
hmm but I think this is just a list of Apps we recommend are good for cost optimization. Potentially they could manage by themselves WDYT?

njuettner:
OpenCost is pretty much useless, we had feedback from customer. I can change the URL to the upstream repo but we definitely don't want to support it.

Shall I change the URL to upstream then?

njuettner:
@pipo02mix you're right leaving it there 👍🏻 makes sense to mention it. closing the PR

PR #1988, opened 2023-11-21
Replace search

pipo02mix:
### What does this PR do?

It removes the current search box and adds a Search Box widget offered by Inkeep. The new search box has AI Capabilities and have better results in general.

### What does it look like?

General Search Box
Screenshot 2023-11-20 at 15 57 00

Sidecar Search Box
Screenshot 2023-11-20 at 15 56 54

Search regular results
Screenshot 2023-11-20 at 15 57 23

Chat AI capabilities 
Screenshot 2023-11-20 at 15 57 07

We have been upset with the search results offered by our current implementation of ElasticSearch.

Check the style and appearance is good

yes


scripts/update-website-content/overrides.css
```diff
@@ -37,7 +37,7 @@ a {
 
 .header-container-wrapper {
   background: #fff;
-  z-index: 999999;
+  z-index: 999;
```
pipo02mix, line 40: It is needed to avoid header to go above the Search Modal

src/static/css/inkeep.css
```diff
@@ -0,0 +1,4 @@
+.ikp-search-bar-trigger__wrapper--size_compact {
+  width: min(750px,100%);
+  height: min(40px,100%);
+}
```
pipo02mix, line 4: This is for the width on the general search box

src/assets/scripts/base.js
```diff
@@ -27,224 +27,32 @@ function getParameterByName(name) {
...
+/* Initialize Inkeep widget for the Search Box */
+var inkeepWidget = null;
+$(function() {
+  inkeepWidget = Inkeep().embed({
+    componentType: "SearchBar", // required
+    targetElement: document.getElementById("searchInputMobile"), // required
+    properties: {
+      stylesheetUrls: ["/css/inkeep.css"],
+      baseSettings: {
+        apiKey: "97dd6334a2ba9121a71ec3e88216e593670153eba46954b0", // required
```
AndiDog, line 39: Can I assume this is okay to have public, and the widget is limited to only be used on `docs.giantswarm` (and maybe `localhost`)?

AndiDog (changes requested):
Open question.

Plus: we need to document how this works, for example in the README. Or at least a link to the intranet where we will describe the Inkeep setup in private.

src/assets/scripts/base.js
pipo02mix: yeah that is what I asked last week to Nick, it seems there is no problem with that 
https://gigantic.slack.com/archives/C05UY4GNN7M/p1700491629729709?thread_ts=1700490386.730039&cid=C05UY4GNN7M

pipo02mix:
> Plus: we need to document how this works, for example in the README. Or at least a link to the intranet where we will describe the Inkeep setup in private.

Good point, I will include it in README and also pìng WG-ai, I think we could add a new place in our intranet/handbook for documenting all AI initiatives

pipo02mix:
BTW I have checked the blog contents and change/releases are also indexed :)

pipo02mix:
I have just sent you an invite to review together tomorrow

AndiDog (approved):

pipo02mix:
Together with Marian, we found some inconsistencies in the search results. I elaborated a 10-point list to check if the search engine returns good results. Now after some fixes, it is working as expected 

Screenshot 2023-11-30 at 10 11 40
Screenshot 2023-11-30 at 17 08 54
Screenshot 2023-11-30 at 17 09 12

cc/ @marians

marians (approved):
Let's go ahead then :)

PR #1998, opened 2023-11-27
Fix menu style

pipo02mix:
### What does this PR do?

Fix the sidebar menu so in case the first level option is clicked the arrow is change to active and the children are shown

### What does it look like?

Before
Screenshot 2023-11-27 at 11 11 53

After
Screenshot 2023-11-27 at 11 12 36

### DIFF
--- a/src/layouts/partials/menu.html
+++ b/src/layouts/partials/menu.html
@@ -3,27 +3,28 @@
 <section class="nav-wrap">
 	<nav class="acnav" role="navigation">
 		<ul class="acnav__list acnav__list--level1">
+
 			{{ $currentPage := . }}
             {{ range .Site.Menus.main }}
                 {{ $pageForItem1 := $site.GetPage .URL }}
                 {{ if .HasChildren }}
                     <li class="has-children">
-                        <div class="acnav__label{{ if or ($currentPage.HasMenuCurrent "main" .) ($currentPage.IsMenuCurrent "main" . ) }} active{{ end }}"><a href="{{ .URL }}" class="{{ if $currentPage.IsMenuCurrent "main" . }}current{{ end }}">{{ .Name }}</a></div>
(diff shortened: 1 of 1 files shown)


src/layouts/partials/menu.html
```diff
@@ -3,27 +3,28 @@
...
+
 			{{ $currentPage := . }}
             {{ range .Site.Menus.main }}
                 {{ $pageForItem1 := $site.GetPage .URL }}
                 {{ if .HasChildren }}
                     <li class="has-children">
-                        <div class="acnav__label{{ if or ($currentPage.HasMenuCurrent "main" .) ($currentPage.IsMenuCurrent "main" . ) }} active{{ end }}"><a href="{{ .URL }}" class="{{ if $currentPage.IsMenuCurrent "main" . }}current{{ end }}">{{ .Name }}</a></div>
-                        {{ if $currentPage.IsDescendant $pageForItem1 }}
+                        <div class="acnav__label{{ if or ($currentPage.HasMenuCurrent "main" .) (eq $pageForItem1.Name $currentPage.Name) }} active{{ end }}"><a href="{{ .URL }}" class="{{ if $currentPage.IsMenuCurrent "main" . }}current{{ end }}">{{ .Name }}</a></div>
+                        {{ if or ($currentPage.IsDescendant $pageForItem1) (eq $pageForItem1.Name $currentPage.Name) }}
```
pipo02mix, line 13: For some weird reason the function `IsMenuCurrent` does not return right value, so I had to fallback to compare the Name (ID is no available).

marians: Strange indeed. Did you try [HasMenuCurrent](https://gohugo.io/methods/page/hasmenucurrent/) as well?

pipo02mix: yes does not work, returns false 🤷

PR #2003, opened 2023-11-28
Update CircleCI configuration for pushing to container registries

marians:
Towards https://github.com/giantswarm/roadmap/issues/2979

This PR was created through automation by Team Honeybadger, to make sure that the container images built for this repository are available in the right registries.

For that, the CircleCI configuration is modified to use the new [push-to-registries](https://github.com/giantswarm/architect-orb/blob/main/docs/job/push-to-registries.md) job instead of the old `push-to-docker`. This means that there is **only one job for all registry pushes**, and it simplifies the configuration by removing many parameters that are now set automatically or as defaults.

## Notes

- Since the PR is also changing the CircleCI job name to `push-to-registries`, the branch protection rules in this repository will likely require updating before the required checks can be green.
- Dev images are sent to ACR (gsoci.azurecr.io) and Quay currently. If you need dev images elsewhere, you can add the according configuration as described in the [documentation](https://github.com/giantswarm/architect-orb/blob/main/docs/job/push-to-registries.md) to the step configuration.

## To the responsible team

Please,

- Double-check the job dependecies (`requires`).
- Double-check the job `filters`. Especially if this PR replaces several jobs with one, and the previous jobs had different filters, make sure that the filter includes all cases.
- Update the changelog if you think this deserves a mention.
- Approve and merge this PR once it's ready. No need to wait for the author in this case.
- Get this done until 1st of December, so that we can move on with follow-up steps.


marians:
Seems like the `DOCKER_IMAGE_TAG` variable is not accessible to the container-test step. Will have to fix this.

marians:
This PR is now also removing the extra container check, which I didn't manage to get right with the use of push-to-registries (I couldn't access the DOCKER_IMAGE_TAG variable). It wasn't critical anyway IMO.

.circleci/config.yml
```diff
@@ -1,65 +1,28 @@
...
-          version: 20.10.11
-
-      - architect/push-to-docker:
-          image: "quay.io/giantswarm/docs"
-          tag-latest-branch: "main"
-          username_envar: "QUAY_USERNAME"
-          password_envar: "QUAY_PASSWORD"
-
-      - run:
-          name: Launch container for tests
```
AndiDog, line 26: @pipo02mix this test gets removed. Should we keep it?

pipo02mix: I guess we can keep it :) should not be affected by this change

pipo02mix (approved):

PR #2005, opened 2023-11-29
Added docs for Device auth flow in kgs login

vvondruska:
### What does this PR do?

Add documentation for a new `--device-auth` flag in `kubectl-gs login`.

### What does it look like?

The documentation of `kubectl-gs login` hs been extended to include the usage description and examples of the new `--device-auth` flag.

Updating documentation after the device auth implementation in `kubectl-gs`: https://github.com/giantswarm/roadmap/issues/1933

Check if the new parts of the documentation make sense


marians:
FYI: branch protection rules are changed to match with https://github.com/giantswarm/docs/pull/2003

anvddriesch (approved):

LolloneS (approved):
Awesome! Thanks!

src/content/use-the-api/kubectl-gs/login.md
```diff
@@ -121,6 +123,46 @@ For subsequent logins, you can use that shorthand to easily select which install
...
+
+Run the `login` command with the `--device-auth` flag to authenticate using the device flow...
+
+```nohighlight
+kubectl gs login https://api.g8s.example.westeurope.azure.gigantic.io --device-auth
+```
+
+Watch the command's output. It will print the URL of the authentication endpoint:
+
+```nohighlight
```
AndiDog, line 136: If this is meant for headless usage, then the only thing printed to stdout should be the URL, no human hints, so that a script or command line could easily consume it. Or JSON output, for example. If we keep the feature as-is, the flag could be called `--no-open-browser` since the remaining behavior seems to be the same.

PR #2007, opened 2023-11-30
Add team annotation in chart.yaml

MarcelMue:
Towards https://github.com/giantswarm/giantswarm/issues/29088

This is mostly to be compliant with component metrics from Atlas. Uncertain how this will behave in other scenarios currently.

### DIFF
--- a/helm/docs-app/Chart.yaml
+++ b/helm/docs-app/Chart.yaml
@@ -3,4 +3,6 @@ name: docs-app
 appVersion: 0.0.1
 description: Giant Swarm documentation published at https://docs.giantswarm.io/
 home: https://github.com/giantswarm/docs/
+annotations:
+  application.giantswarm.io/team: "sig-docs"
 version: "[[.Version]]"


lyind (approved):
It did a search over our `giantswarm` org to identify how the labels are used:

- How the time is specified varies:
  - `team-honeybadger` VS `honeybadger`
  - `tinkerers` VS `team-tinkerers`
  - *the correct way is defined in `app-exporter` and should be just the team name:* https://github.com/giantswarm/app-exporter/blob/7f5e8de74c32c6ea1d08bc18af6fd8c78d7de29c/service/collector/app_test.go#L188
- `app-exporter` and the prometheus alerting are the only current users I can identify
- `app-exporter` may run into trouble with SIGs and also the misspelled team names, later. (It has a concept of retired teams and "team mappings" which may just not map, then.)

**NOTE:** I did approve this PR, because app-exporter could easily strip the `team-` and detect the `SIG`. 😸

PR #2008, opened 2023-11-30
Update policy enforcement docs with fresh Policy API info

stone-z:
### What does this PR do?

Mostly, adds some missing links and a sample GS PolicyException

### What does it look like?

This is one of several updates with requested improvements to these docs to support the PSS transition.

Front matter is still current, will bump the review on next update.


stone-z:
CI failure is unrelated to the content change

LolloneS (approved):

PR #2017, opened 2023-12-12
Add note about selecting AWS account when creating CAPA clusters

fiunchinho:
### What does this PR do?
Towards https://github.com/giantswarm/giantswarm/issues/29249

When creating clusters using CAPA, one of the most common questions will be "How do I select the AWS account where resources will be created?". So to ease the cluster creation, we add a mention to this in the getting started page, and link to the page with more details about configuring AWS accounts.


calvix (approved):

LolloneS (approved):
LGTM, thanks!

PR #2018, opened 2023-12-12
Add a note mentioning app name length limits

weseven:
### What does this PR do?
Add a note mentioning app name length limits

This comes from the issues arising when increasing cluster name length to 20 characters: in CAPI clusters the cluster name is used as prefix for the App CRs name, so the effective app name length must be <32 characters for a successful deploy.


LolloneS (approved):
LGTM but please fix `last_review_date`, it should be `last_review_date: 2023-12-12``

weseven:
> LGTM but please fix `last_review_date`, it should be `last_review_date: 2023-12-12`

Thanks, updated in the last commit!

PR #2028, opened 2023-12-21
Mention instance refresh warmup settings for CAPA, add general advice

AndiDog:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/2217, as suggested by @fiunchinho in https://github.com/giantswarm/cluster-aws/pull/457.

yes


fiunchinho:
The text looks fine. But I was wondering about the 5 minutes warm up. Won't this be a problem on big clusters? We have customers that have clusters with hundreds of nodes. Wouldn't the rolling upgrade take many hours to complete?

AndiDog:
> The text looks fine. But I was wondering about the 5 minutes warm up. Won't this be a problem on big clusters? We have customers that have clusters with hundreds of nodes. Wouldn't the rolling upgrade take many hours to complete?

From how I understand ASGs, they don't replace one-by-one, but rather based on "minimum healthy percentage". It's 90% by default, I believe, but of course at least one node is replaced at once. For large clusters, that would mean that more than one node would get replaced at once.

src/content/advanced/cluster-management/node-pools-capi/index.md
```diff
@@ -216,12 +216,25 @@ If you still want to change the name of a node pool, we recommend adding a new n
...
+
+{{< /tab >}}
+{{< tab id="nodepool-update-capa" for-impl="capa_ec2" >}}
+
+During a cluster upgrade, it can be necessary to create new EC2 instances (we call this "rolling nodes"). That only happens if anything changed in the node configuration, such as configuration files or the AMI image (newer version of Kubernetes or Flatcar Linux). For such planned node replacements, we are using [instance warmup settings](https://github.com/search?q=repo%3Agiantswarm%2Fcluster-aws%20refreshPreferences&type=code) to ensure that AWS does not replace the old nodes too quickly all at once, but rather in steps so a human could still intervene if something goes wrong (e.g. roll back to previous version). For a small node pool, that would mean that one node would be replaced every 5 minutes.
```
fiunchinho, line 229: What do you think about this?

```suggestion
For a small node pool, that would mean that one node would be replaced every 5 minutes. For bigger node pools, small groups of nodes would be replaced every 5 minutes.
```

fiunchinho (approved):

PR #2040, opened 2024-01-10
Reduce container image size

marians:
### What does this PR do?

This PR changes the container image and nginx config so that only compressed HTML files are stored. As a result, the image size goes from 949 MB down to 159 MB.

https://github.com/giantswarm/roadmap/issues/3033


AndiDog (approved):
Can't tell from code starring whether this could work. If you tested it, or will watch the successful deployment to be ready for rollback, I'm fine.

marians:
I have indeed tested it locally and will watch the rollout

lyind:
This is awesome, going to steal this for giantswarm/intranet and giantswarm/handbook.

PR #2046, opened 2024-01-17
Minor improvements to Policy API documentation

stone-z:
### What does this PR do?

Adds links which didn't work during the initial PR, and revisits some of the previously rushed content.

### What does it look like?

Toward https://github.com/giantswarm/giantswarm/issues/28243

Won't bump now as we expect larger changes coming soon.


Strigix:
Approved, would you reckon, that in future, we might have to document for customers to read which already existing options we've considered? With a description why we choose to not go with said already existing options?

Strigix (approved):

pipo02mix (approved):

pipo02mix:
merging here

PR #2052, opened 2024-01-18
Update kubectl gs get releases

mproffitt:
Change of review date only - no content changes


marians (approved):

marians:
I'm thinking that by now a note regarding vintage only would make sense here.

mproffitt:
Damn - that comment came in seconds before automerge - It would be correct. I'll do a new PR for that.

PR #2105, opened 2024-01-31
Improve PR template

AndiDog:
Make it simpler so PR authors actually fill it in, and hint at `make dev` for local testing since that can avoid styling and rendering problems.


LolloneS (approved):

.github/pull_request_template.md
```diff
@@ -1,17 +1,8 @@
...
-### Any background context you can provide?
-
-(Please link public issues or summarize if not public.)
-
-### What is needed from the reviewers?
-
-### Have you maintained the front matter?
-
-(Please bump the last_review_date in case this qualifies as a review for an entire page. Provide user_questions which should be answered in the page. Provide a meaningful description.)
+- [ ] (Content changes only) Ran `make dev` to test content changes locally.
```
marians, line 7: This looks like we expect everybody to test changes locally. Do we do that?

I wouldn't recommend putting something optional into the checklist.

.github/pull_request_template.md
```diff
@@ -1,17 +1,8 @@
...
+- [ ] (Content changes only) Bumped `last_review_date` in the front matter header because I reviewed the entire page.
```
marians, line 8: Again, it's fuzzy whether we want this on every docs PR now. I recommend against it and so would not make it part of a checklist.

AndiDog: Yes, both items are optional. Not great, I know. At least `make dev` is _recommended_, and the review date bump is _conditional_. If we don't mention it here, nobody will take care of that or know about it. Which alternatives do we have here? One idea is to make it a bullet point list "Things to check/remember" instead of items to be ticked off.

marians: > One idea is to make it a bullet point list "Things to check/remember" instead of items to be ticked off.

Yes, I like that better.

.github/pull_request_template.md
pipo02mix: IMO it is a good reminder in case you have modified a content file (which are the 95% of the cases)

pipo02mix (approved):
LGTM (when Marian agrees)

.github/pull_request_template.md
AndiDog: Changed to bullet points

marians (approved):
Fine, thanks

PR #2108, opened 2024-02-01
Dockerfile: avoid exec overhead in compression step

lyind:
### Summary

While adapting the `gzip` compression step for Intranet and Handbook, it was noticed that the build stage took longer than expected.

A large part of that time is due to `gzip` process launch overhead for each of the individual files.

Make this scale better by using as few processes as possible by means of `find | xargs gzip`. 🥬🍃

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,8 @@ RUN hugo \
 # (creates a copy and leaves the uncompressed version in place)
 RUN find /public \
   -type f -regextype posix-extended \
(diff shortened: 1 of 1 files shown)


marians:
What's the time difference? I checked the CircleCI job runs and can't notice a difference.

lyind:
I didn't measure, just ran locally and found it to be a bit quicker (as it should be).

Scales with the number of files compressed and depends on the process launch overhead of your individual machine.

marians (approved):

PR #2114, opened 2024-02-08
Restrict security settings, fix ServiceAccount

marians:
Towards https://github.com/giantswarm/giantswarm/issues/29470

The main changes in this PR:

- Restrict the pod/container security spec so that the PSS PolicyException resource is not needed
- Render the ServiceAccount resource independent from the PSP.

Other changes:

- Set the registry name in values.yaml
- Change the imagePullPolicy from ` Always` to `IfNotPresent` to avoid obsolete pulls


stone-z:
This is great, thanks!

> Change the imagePullPolicy from  `Always` to `IfNotPresent` to avoid obsolete pulls

Doesn't this work the other way around? IfNotPresent could result in an old image with the same tag being used

marians:
With the release automation in this repo, it normally shouldn't happen that an image tag is re-used.

With `IfNotPresent` we save some pulls in the case that containers get launched on the same node that ran this version before. Which admittedly should also be not very frequent.

stone-z:
Saving pulls makes sense, the avoiding stale pulls didn't (I understood that to mean "avoid re-using an obsolete image"). But all good from my side, just checking this was the intended behavior

PR #2130, opened 2024-02-21
Add CRD reference for PolicyException, PolicyExceptionDraft

marians:

As we want documentation for our CRDs to be accessible easily, I'm adding those for the Policy API.

Besides, the crd-docs-generator gets updated to the latest version.


stone-z (approved):
Thank you! I previously tried to figure out how to do this and couldn't -- I thought only the `apiextensions` CRDs were supported.

marians:
We are by now in fact quite flexible regarding where to read CRDs from. :)

PR #2131, opened 2024-02-21
Update crd-docs-generator to v0.11.1, change alias

marians:
This PR updates crd-docs-generator to the latest version. Also in the CRD ref template, the alias is updated to add an alias for the URI before the vintage move.

It looks like annotations are being removed (or rather: not added) to the generated output, so I'm checking why.

### DIFF
--- a/scripts/update-crd-reference/crd.template
+++ b/scripts/update-crd-reference/crd.template
@@ -50,7 +50,7 @@ owner:
   - {{ . -}}
 {{- end }}
 aliases:
-  - /reference/cp-k8s-api/{{ .NamePlural }}.{{ .Group }}/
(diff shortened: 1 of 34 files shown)


scripts/update-crd-reference/crd.template
```diff
@@ -50,7 +50,7 @@ owner:
   - {{ . -}}
 {{- end }}
 aliases:
-  - /reference/cp-k8s-api/{{ .NamePlural }}.{{ .Group }}/
+  - /use-the-api/management-api/crd/{{ .NamePlural }}.{{ .Group }}/
```
pipo02mix, line 53: ❤️

marians:
The aliases change is now isolated in https://github.com/giantswarm/docs/pull/2150

PR #2134, opened 2024-02-22
Force adding cluster apps into the docs

pipo02mix:

Towards https://github.com/giantswarm/giantswarm/issues/29628

Job is not working properly, taking a look a part, for now I push this manually 

- If you made content changes


pipo02mix:
> This is part of the "old" subpath for the docs. Do we really want to do this already? The content is for CAPI.

It will be part of the new docs, but I don't know where yet. It is part of the migration epic :)

PR #2153, opened 2024-03-06
Do not render last_review_date in generated helm charts docs

marians:
Goal: Avoid changing chart content every day, keep diffs for generated content as small as possible.


marians:
FYI @pipo02mix

AndiDog (approved):

PR #2154, opened 2024-03-06
Fix changes path, ignore charts path

marians:
Follow-up for https://github.com/giantswarm/docs/pull/2153 where I made some mistakes.


lyind (approved):
We are also using this script in giantswarm/giantswarm (Intranet).

In a distant future, we might want to make the paths configurable outside the script, maybe even move the script itself into a separate GitHub action to avoid having to copy it.

Otherwise LGTM

PR #2172, opened 2024-04-02
Fix check links and make it simpler

pipo02mix:

It split the check link validation for doc section making it easier to track errors. Also it reduces the checks in `changes/` section since we don't want to fix external URLs for old changelog entries

- If you made content changes

### DIFF
--- a/.github/workflows/check-links-in-prod.yaml
+++ b/.github/workflows/check-links-in-prod.yaml
@@ -73,23 +73,23 @@ jobs:
       - name: Report errors
         run: |
           if [[ $failed1 == "true" ]]; then
-            echo "There has been some errors in overview checks"
+            echo "There has been some errors in overview checks, please check the step."
           fi
           if [[ $failed2 == "true" ]]; then
-            echo "There has been some errors in getting started checks"
+            echo "There has been some errors in getting started checks, please check the step."
           fi
           if [[ $failed3 == "true" ]]; then
-            echo "There has been some errors in tutorials checks"
+            echo "There has been some errors in tutorials checks, please check the step."
           fi
           if [[ $failed4 == "true" ]]; then
-            echo "There has been some errors in reference checks"
+            echo "There has been some errors in reference checks, please check the step."
           fi
           if [[ $failed5 == "true" ]]; then
-            echo "There has been some errors in support checks"
+            echo "There has been some errors in support checks, please check the step."
           fi
           if [[ $failed6 "true" ]]; then
-            echo "There has been some errors in changelogs checks"
+            echo "There has been some errors in changelogs checks, please check the step."
           fi
-          if [[ $failed1 == "true" ]] || [[ $failed1 }}" == "true" ]] || [[ $failed1 == "true" ]] || [[ $failed1 == "true" ]] || [[ $failed1 == "true" ]] || [[ $failed1 "true" ]]; then
+          if [[ $failed1 == "true" ]] || [[ $failed2  == "true" ]] || [[ $failed3 == "true" ]] || [[ $failed4 == "true" ]] || [[ $failed5 == "true" ]] || [[ $failed6 "true" ]]; then
             exit 1
           fi
--- a/.github/workflows/check-links-in-prod.yaml
+++ b/.github/workflows/check-links-in-prod.yaml
@@ -18,7 +18,7 @@ jobs:
             ghcr.io/linkchecker/linkchecker:latest \
             https://docs.giantswarm.io/overview/ \
             --check-extern \
-            -t 1 -r 2 || echo "::set-output name=failed::true"
+            -t 1 -r 2 || echo "failed=true" >> "$GITHUB_OUTPUT"
         continue-on-error: true
 
       - name: check-links-in-getting-started-pages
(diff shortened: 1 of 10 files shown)


.github/workflows/check-links-in-prod.yaml
```diff
@@ -12,15 +12,54 @@ jobs:
...
+        run: |
+          docker run --rm --name linkchecker \
+            ghcr.io/linkchecker/linkchecker:latest \
+            https://docs.giantswarm.io/changes \
+            -t 1 -r 2 \
+            --ignore-url="^https://github.com/giantswarm/docs/.*" \
+            --ignore-url="^https://.*example\.com/.*" \
+            --ignore-url="^https://my-org\.github\.com/.*" \
+            --ignore-url="^https://github\.com/giantswarm/giantswarm/.*" \
+            --ignore-url=".*gigantic\.io.*"
```
pipo02mix, line 65: This makes it easier to consume the output and does not fall into the whole of deep validation (we do have not control over when the old changelog points to a non-existing URL)

marians:
It seems like this is not working as intended. I ran it in https://github.com/giantswarm/docs/actions/runs/8539886897/job/23395648737 and the first check starts on the /overview page, but recurses into all other pages. So we end up checking the entire site multiple times.

.github/workflows/check-links-in-prod.yaml
```diff
@@ -12,15 +12,54 @@ jobs:
...
+            -t 1
+
+      - name: Check links in reference pages
+        run: |
+          docker run --rm --name linkchecker \
+            ghcr.io/linkchecker/linkchecker:latest \
+            https://docs.giantswarm.io/reference \
+            --check-extern \
+            -t 1
+
```
pipo02mix, line 46:
```suggestion
            https://docs.giantswarm.io/overview/ \
            --check-extern \
            -t 1 -r 2

      - name: Check links in getting-started pages
        run: |
          docker run --rm --name linkchecker \
            ghcr.io/linkchecker/linkchecker:latest \
            https://docs.giantswarm.io/getting-started/ \
            --check-extern \
            -t 1-r 2

      - name: Check links in tutorials pages
        run: |
          docker run --rm --name linkchecker \
            ghcr.io/linkchecker/linkchecker:latest \
            https://docs.giantswarm.io/tutorials/ \
            --check-extern \
            -t 1 -r 2

      - name: Check links in reference pages
        run: |
          docker run --rm --name linkchecker \
            ghcr.io/linkchecker/linkchecker:latest \
            https://docs.giantswarm.io/reference/ \
            --check-extern \
            -t 1 -r 2

```

.github/workflows/check-links-in-prod.yaml
```diff
@@ -12,15 +12,54 @@ jobs:
...
+            https://docs.giantswarm.io/support \
+            --check-extern \
+            -t 1
+
+      - name: Check links in changelogs
+        run: |
+          docker run --rm --name linkchecker \
+            ghcr.io/linkchecker/linkchecker:latest \
+            https://docs.giantswarm.io/changes \
+            -t 1 -r 2 \
```
pipo02mix, line 60:
```suggestion
      - name: Check links in support pages
        run: |
          docker run --rm --name linkchecker \
            ghcr.io/linkchecker/linkchecker:latest \
            https://docs.giantswarm.io/support/ \
            --check-extern \
            -t 1 -r 2

      - name: Check links in changelogs
        run: |
          docker run --rm --name linkchecker \
            ghcr.io/linkchecker/linkchecker:latest \
            https://docs.giantswarm.io/changes/ \
            -t 1 -r 2 \
```

pipo02mix:
[Fixed](https://github.com/giantswarm/docs/actions/runs/8554223183/job/23439085880). I will try to pass the tests too

.github/workflows/check-links-in-prod.yaml
```diff
@@ -12,15 +12,54 @@ jobs:
...
+            --check-extern \
+            -t 1 -r 2
+
+      - name: Check links in getting-started pages
+        run: |
+          docker run --rm --name linkchecker \
+            ghcr.io/linkchecker/linkchecker:latest \
+            https://docs.giantswarm.io/getting-started/ \
+            --check-extern \
+            -t 1-r 2
```
marians, line 29:
```suggestion
            -t 1 -r 2
```

pipo02mix:
sadly it runs against the current online docs.giantswarm.io page so I cannot demonstrate the changes fix the linter :)

pipo02mix:
@marians can you take a look?

marians:
I just ran the workflow here: https://github.com/giantswarm/docs/actions/runs/8615042101/job/23609717739

The first step finds broken links and exits with non-zero exit code. As a result, the workflow does not continue after the step, so other links are not checked.

pipo02mix:
yeah that is a problem, maybe I can cache all errors and only stop if the array is not empty, let me try

pipo02mix:
now should be okish @marians PTAL

marians:
Looks like we are getting there!

`::set-output` is deprecated and will be removed. See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ for info on what to use instead.

marians:
Not sure where I saw `::set-output`. Probably looked at an outdated diff.

marians (approved):
LGTM so far. Ideally let's switch to something else than `::set-output

PR #2180, opened 2024-04-16
Include a template for overview

pipo02mix:

Part of https://github.com/giantswarm/giantswarm/issues/30016

- If you made content changes


src/content/overview/_template.md.tpl
```diff
@@ -0,0 +1,24 @@
+---
+title: Short title for this page, it is shown as title in upper section and menu
+description: At least 70 characters that describe the purpose of this page.
+weight: from 10 to 100. Lower is higher in the menu.
```
AndiDog, line 4:
```suggestion
weight: A number from 10 to 100 defining the order in the menu. Lower number = top of menu, higher number = bottom of menu. Use multiples of 10 (= 20, 30, 40, ...) when adding a page so that we can later easily put things in between, if necessary, without changing the weight in several files.
```

src/content/overview/_template.md.tpl
```diff
@@ -0,0 +1,24 @@
+---
+title: Short title for this page, it is shown as title in upper section and menu
+description: At least 70 characters that describe the purpose of this page.
```
AndiDog, line 3:
```suggestion
description: At least 70 characters that describe the purpose of this page. Do not repeat the `title`, and do not use fill words such as "Overview of" or "A short explanation how we".
```

AndiDog (approved):
LGTM, please merge with or without my suggestions – they're minor since this is a template

PR #2189, opened 2024-04-22
Add prose linting

marians:
Towards https://github.com/giantswarm/roadmap/issues/3406

This adds prose linting for content.

This PR **does't** make all content compliant.

Run `make lint-prose` or, for slightly faster execution, run `vale --glob '!{src/content/vintage/**,src/content/changes/**}' src/content`

You can find one annotation on the commit https://github.com/giantswarm/docs/pull/2189/commits/2f625c523d17dcfc97f65361d71f8a206e09f11d.

## Changes

- The `make lint` target previously linted only Markdown. Now it validates markdown, front matter, and prose.
- The make target `make lint-prose` is added.
- The make target `make lint-markdown` is added (previously `make lint`).
- The `.vale` folder is added with vale styles and vocabulary
- The `.vale.ini` file is added with vale configuration
- Add workflow/action to run the linting for PRs
- The file `src/content/reference/fleet-management/clusters/introduction_cluster_api.md` is updated to comply with most rules, except for one: it uses `K8s`, which is on our list of terms to avoid.

## Regarding the workflow added

I couldn't get the `vale-action` to work, so I decided to download and execute the binaries for vale and reviewdog directly.

The workflow only reports errors for the files touched.

The implementation can be improved to execute Vale only on the modified files.


mcharriere (approved):
really nice :)

.vale.ini
```diff
@@ -0,0 +1,12 @@
...
+Vocab = docs
+
+Packages = Hugo, Microsoft
+
+[*.md]
+MinAlertLevel = suggestion
+BasedOnStyles = Vale, Microsoft
+
+# Microsoft.We disallows the use of "we" or "us". We are using this in many places for now.
+Microsoft.We = NO
```
AndiDog, line 12: Well, in the future this should indeed go away. We mix "we"/"us" vs. "you" way too much in our tutorials. Instead, it should be "Giant Swarm [platform]" (instead of "we") vs. "you"/"customer".

.vale/styles/Microsoft/AMPM.yml
```diff
@@ -0,0 +1,9 @@
+extends: existence
```
AndiDog, line 1: Do we need to _copy_ all the Microsoft rules here, or could we instead reference a specific version of theirs somehow?

If not possible, we should add a README description about how to update/sync these files, please.

AndiDog (commented):
Great addition

.vale.ini
marians: I wanted to wait with that one, since I had mixed feelings and wanted some debate before enabling it. I always liked the personal touch of "we" (Giant Swarm) and "you" (the customer). But I'm not sure it comes across as professional, and also I would like to figure out what the alternatives would be.

.vale/styles/Microsoft/AMPM.yml
marians: The copying is a consequence of executing `vale sync`. Without these files, these third party rules cannot be used for linting.

I will add some info to the README.

mcharriere: I'm wondering if shouldn't run `vale sync` on CI. If we don't, CI always checks what we have in the repo. If we do, CI runs something different, that might lead to some drift.

I'm for removing the `vale sync` and maybe look for some automation for updates. Although I don't know how often we need to update that.

marians: Good point

marians: Done

PR #2191, opened 2024-04-26
Adapt content to style guide

marians:
This PR adapts existing content to our style guide, so that we get a consistent starting point with minimal linter warnings.


pipo02mix (approved):
LGTM maybe closing the app name between quotes will solve it (though it means we need to do it for all)

marians:
> LGTM maybe closing the app name between quotes will solve it (though it means we need to do it for all)

Indeed, backtick comments would work, but I think it's the wrong thing to do, in terms of style/formatting.

PR #2198, opened 2024-05-02
Draft tutorial for Cilium ENI IPAM mode on AWS

AndiDog:

Hidden draft in the revamped docs. I want to get an understanding if the outline roughly makes sense. The actual content will be filled in another PR, but you can also comment on wording already if you want.

- If you made content changes

    - Run `make dev` to render and proofread content changes locally. => ❌ currently failing and my article is only a Markdown comment


src/content/tutorials/aws-cilium-eni-mode.md
```diff
@@ -0,0 +1,128 @@
...
+---
+
+<!--
+
+A workload cluster can be configured to choose pod IPs from an AWS-allocated IP range (CIDR). In this mode, the Cilium CNI [allocates ENIs (Elastic Network Interfaces) and pod IPs](https://docs.cilium.io/en/latest/network/concepts/ipam/eni/).
+
+## Advantages
+
+- The separate CIDR, and the separate subnets created for the pods, can be used for handling pod traffic separately. For example, firewalling or peering can be done only for those IPs.
+- Pod traffic is not translated by NAT inside the cluster. If you have existing security tooling, for instance, that is not adapted to Kubernetes, it can show the pod IPs as if they were a regular server on your network. On clusters not running in Cilium ENI mode, you would typically see the node IP instead of the pod IP. Mind that modern observability tooling such as Prometheus integrates well with Kubernetes so that pod names and IPs can easily be exposed, for example as metric labels.
```
calvix, line 20: in pod-to-pod traffic, you will always see the source pod IP in the destination pod

and between nodes NAT is not used as calico send traffic via VXLAN tunnels

So this line seems little bit inaccurate, maybe I would simplify the Advantages as there aren't that many TBH, and if customer wants it they already know the advantages and disadvantages. 

I would you to list this:
* Pods get directly assigned IPs that belong to AWS Subnet and are assigned to AWS Security group. 
* Pod IP can be visible in peered VPC or traffic going thru Transit Gateway.

AndiDog: Thanks, I agree what I wrote is inaccurate and probably too much for a tutorial where customers are already decided. I reduced the advantages to pretty much what you suggested.

calvix (approved):

PR #2205, opened 2024-05-07
Remove unecessary intro from cluster charts

pipo02mix:

Towards https://github.com/giantswarm/giantswarm/issues/29503

We want to control the layout of the cluster app charts, now all cluster app repos have different README content that I would like to avoid

- If you made content changes


scripts/update-helm-chart-reference/chart.template
```diff
@@ -1,5 +1,5 @@
 ---
-title: {{ .Title | title }} chart reference
+title: {{ .Title }} chart reference
```
pipo02mix, line 2: This adds weird case to composed chart names like `cluster-aws`

AndiDog (approved):

PR #2261, opened 2024-07-24
Add Azure releases to docs

yulianedyalkova:

Now that we have release [v25.0.0](https://github.com/giantswarm/releases/tree/master/azure/v25.0.0) for Azure, we also need to start generating docs for it.

- If you made content changes

    - Run `make lint dev` to render and proofread content changes locally.


njuettner (approved):

T-Kukawka (approved):

T-Kukawka:
cc @pipo02mix

pipo02mix:
wait, but is this CAPZ? Why don't we name the folder `capz`? That way, we follow the convention `capa`, `capv`,...

pipo02mix:
@yulianedyalkova ?

yulianedyalkova:
@pipo02mix I was afk :) The longterm plan is to use the names of the providers rather than capa/capz/etc. CAPA was used because of conflicts with vintage but even there we want to eventually go back to just using aws.

pipo02mix:
and what happens with AKS and pure Azure? or aws and eks?

yulianedyalkova:
That is a conversation that started in [this rfc](https://github.com/giantswarm/rfc/pull/86) but there is no final decision yet. Azure vs capz is certain though as capz is just an implementational detail that shouldn't reach customers. So can we move forward with this PR as it's currently blocking us from announcing the azure releases we already have out?

marians (approved):

PR #2264, opened 2024-08-04
Semi-automatic update of generated content

taylorbot:
This PR has been created automatically by the 'Update generated content' Github workflow
to keep

- Changes and releases
- CLI versions
- Cluster Apps
- CRD pages
- Website header and footer

up-to-date.

This PR is opened or updated every weekday morning. On Fridays, it is merged by 'merge-autogenerated-content' workflow. In case of being triggered manually you need to merge it manually too in order to have the content published.


yulianedyalkova:
DO NOT MERGE!

Currently this PR has two issues:
* `azure` needs to get added [here](https://github.com/giantswarm/docs/blob/aa6c5dfcf149567424338de1bf3c69d7eb0f0436/scripts/aggregate-changelogs/script.py#L241) otherwise we end up with the provider name being added twice in the path name (`/changes/tenant-cluster-releases-azure/releases/azure-azure-25.0.0/`)
* it's publishing all the deprecated azure releases which I assume we don't want

yulianedyalkova (changes requested):
DO NOT MERGE!

Currently this PR has two issues:
* `azure` needs to get added [here](https://github.com/giantswarm/docs/blob/aa6c5dfcf149567424338de1bf3c69d7eb0f0436/scripts/aggregate-changelogs/script.py#L241) otherwise we end up with the provider name being added twice in the path name (`/changes/tenant-cluster-releases-azure/releases/azure-azure-25.0.0/`)
* it's publishing all the deprecated azure releases which I assume we don't want

PR #2348, opened 2024-11-27
sidebarSearchForm: set absolute width/height to prevent reflow

lyind:
### Summary

Set `width` and `height` to empty search button container to avoid ugly visible reflow of sidebar menu during navigation.

@marians Not sure if this is the correct or best way to do it, but it leads to the desired result on a desktop browser.

### Note

Eyesore originally detected while reviewing: https://github.com/giantswarm/docs/pull/2346

### DIFF
--- a/src/assets/styles/base.sass
+++ b/src/assets/styles/base.sass
@@ -76,6 +76,9 @@ abbr[title]
 #sidebarSearchForm
   font-size: 14px
   margin: 28px 0 28px auto
+  // match Inkeep's dimensions to avoid visible reflow
+  width: 262.5px
+  height: 40px
 
   input
     width: 160px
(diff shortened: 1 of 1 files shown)


src/assets/styles/base.sass
```diff
@@ -76,6 +76,9 @@ abbr[title]
 #sidebarSearchForm
   font-size: 14px
   margin: 28px 0 28px auto
+  // match Inkeep's dimensions to avoid visible reflow
+  width: 262.5px
+  height: 40px
 
   input
```
lyind, line 83: This `input` element doesn't seem to exist anymore, with the Inkeep based search. The specific style for the `button` element may also be unnecessary now.

Shall we remove it or leave it present, in case we switch back to another search form (with a real input field)?

pipo02mix (approved):

PR #2358, opened 2024-11-28
Use title instead of linkTitle in list layout

marians:
In general, I think, we should use the long title form as the header in list templates.

The short title form (`linkTitle`) should only be used in the menu and breadcrumb, where space is heavily constrained.

This PR makes the according switch. I expect that we want to tweak some titles as a consequence, as they might be too long.

Before:

image

After:

image


pipo02mix (approved):
LGTM

PR #2359, opened 2024-11-29
Actions maintenance

marians:


scripts/validate-front-matter/script.py
```diff
@@ -8,6 +8,7 @@
 try:
     from yaml import CLoader as Loader
 except ImportError:
+    print("Warning: Using slower Python YAML loader. Install LibYAML for faster performance.")
```
marians, line 11: To be able to check whether we have the LibYAML dependency installed. Currently we do, but on a future runner we may not.

.github/workflows/validate.yaml
```diff
@@ -9,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Install dependencies
-        run: pip3 install click colored==1.4.4
+        run: pip3 install click colored
```
marians, line 18: No need to keep the version pinned here

lyind (approved):
𝕃𝔾𝕋𝕄

PR #2360, opened 2024-11-29
CI: Report moved, renamed, deleted files

marians:
The idea here is to report on files that got moved/renamed or deleted, to remind users to check links or update aliases.

If files are renamed/moved or deleted, a comment will appear. Preview:

image

If a later commit reverts these changes, and no more files are renamed or deleted, the comment is removed.

While at it, I'm doing some improvements:

- Replace giantswarm/annotations-action with yuzutech/annotations-action (the former was a fork anyway)
- Fix the call of markdown linting (`make lint` -> `make lint-markdown`) to avoid the attempt to also run `vale` )which is happening in a separate workflow already.
- Merge some of the workflow steps, to make it easier to digest.


pipo02mix (approved):
LGTM!

PR #2380, opened 2024-12-04
ci config: migrate to app-build-suite

lyind:
Towards https://github.com/giantswarm/giantswarm/issues/31101#issuecomment-2470434482

### Summary

Switch project to use `app-build-suite`.


helm/docs-app/values.yaml
```diff
@@ -1,10 +1,5 @@
 name: docs-app
 namespace: docs
-image:
```
uvegla, line 3: With PSP removed, you can also remove this block from below:

```yaml
global:
  podSecurityStandards:
    enforced: false
```

Then regenerate the values schema.

uvegla (approved):

helm/docs-app/values.yaml
lyind: done

PR #2381, opened 2024-12-04
CI: build site in Hugo, bring warnings to the PR as comment

marians:
If warnings are yielded on `hugo build`, they will be commented in the PR.

Preview:

image

If errors are found, they will fail the validate job.


lyind:
Wish this could use the regular build process.

Unfortunately that runs on Circle CI, but maybe one could still access and parse the hugo build output from there?

marians:
> Wish this could use the regular build process.
> 
> Unfortunately that runs on Circle CI, but maybe one could still access and parse the hugo build output from there?

Nut sure. But it would be a lot more complicated that way.

lyind (approved):
A bit of duplication, but for a good cause.

LGTM

pipo02mix:
nice

PR #2383, opened 2024-12-04
Fix most duplicate path warnings

marians:
Before this change, `hugo build` would yield this warning:

```
[...]

PR #2384, opened 2024-12-04
Fix path duplicates in CRD pages

marians:
Continuing from https://github.com/giantswarm/docs/pull/2383

This PR mitigates the problem especially for CRD pages.


lyind (approved):
LGTM

PR #2388, opened 2024-12-09
Remove link check for changes section

marians:
This removes the link checking for the Changes and Releases sections, as it's not realistic that we maintain outgoing links in changelogs.


marians:
Obsolete

PR #2413, opened 2024-12-12
Mitigate duplicate menu entry warnings for cluster-app chart pages

marians:
Like in

https://github.com/giantswarm/docs/pull/2412#issuecomment-2539511544

### DIFF
--- a/scripts/update-helm-chart-reference/main.sh
+++ b/scripts/update-helm-chart-reference/main.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 HELM_CHART_DOCS_GENERATOR_VERSION=0.2.0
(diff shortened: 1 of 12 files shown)


scripts/update-helm-chart-reference/main.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 HELM_CHART_DOCS_GENERATOR_VERSION=0.2.0
-DESTINATION=src/content/vintage/use-the-api/management-api/cluster-apps
+
+DESTINATION=src/content/reference/platform-api/cluster-apps
```
marians, line 5: We want new generated content to land in `/reference`.

We don't want the pages in `/vintage` to get updated any more.

pipo02mix (approved):
LGTM

PR #2424, opened 2024-12-13
Some maintenance for content owned by honeybadger

marians:
- Update tutorials > container registries front matter
- Fix headline hierarchy of the above page
- Rename files to force rendering as article
- Rename some files to use dash instead of underscore
- Delete an obsolete article about shared installations that was hidden in the /meta section


pipo02mix (approved):
nice!

PR #2432, opened 2024-12-16
Update links to Rest API docs

marians:
Two benefits:

- No redirect needed, users land on the final URL immediately.
- This will make maintenance easier, as it will be easier to find internal links that do not use the `relref` shortcode.


AndiDog (approved):
That's really outdated REST API stuff, maybe leave it or remove it?

marians:
There is a reason to do this

> This will make maintenance easier, as it will be easier to find internal links that do not use the relref shortcode.

PR #2433, opened 2024-12-16
Enable Renovate to update CRD versions

marians:
If this works, it will enable Renovate to update CRD versions in the configuration of `scripts/update-crd-reference`. With that, the custom script `update_config.sh` becomes obsolete.

A check is added to verify that `make update-crd-reference` runs without error.


pipo02mix (approved):
LGTM

PR #2450, opened 2025-01-17
Removing all references to api-spec

marians:
Removes all pointers to api-spec.

Part of a larger removal of REST API related content.

This also adapts the validate workflow, which failed at the wrong place without the adaptation. The reason I think is that ubuntu-latest is now 24.04 and there are some shell tool changes.


pipo02mix (approved):
👏 LGTM

PR #2453, opened 2025-01-17
Remove vintage getting-started content: setup of cloud provider accounts

marians:
Removes the entire sub-section https://docs.giantswarm.io/vintage/getting-started/cloud-provider-accounts/ as we are not onboarding customers to vintage any more.


pipo02mix (approved):
nice

PR #2456, opened 2025-01-21
Fix layout problems

marians:
Fixes the problem that the pages under https://docs.giantswarm.io/getting-started/prepare-your-provider-infrastructure/ use the list page layout


marians:
Merging without review

PR #2513, opened 2025-03-06
Align doc categories with product areas

pipo02mix:

- If it's one of your first contributions, make sure you've read the [Contributing Guidelines](https://handbook.giantswarm.io/docs/content/docs-guide).
- Bump `last_review_date` in the front matter header of the pages you've touched.


teemow (approved):

stone-z (approved):
This is fine with me as-is but I had a question:
I might have missed it - is this based on a new "product areas" concept? This currently seems more aligned by team ownership, so if there's a broader product area concept I would like to go refine that as well. I think it might be good to define more granular pieces of the platform, and/or to support multiple areas for a single item.

e.g. is `cert-manager` the same Security as `rbac-operator`? `cloudflared`, `cnpg`, and `cert-manager` are all rather different, and it would be weird to me as a customer if `external-dns` and `cilium` are Connectivity, but `cert-manager` and `cloudflared` are not.

pipo02mix:
> I might have missed it - is this based on a new "product areas" concept?

No, it comes from the docs revamp. When I structured the new docs hub, I wanted to group articles in areas. There was already a taxonomy; I have just tried to extend it and make it solid (using the same terms as upstream/community). I have no intention of establishing any categories; I would rather that the product define those, and I have just put them in the docs. Should I bring it to SIG product?

stone-z:
> When I structured the new docs hub, I wanted to group articles in areas. There was already a taxonomy; I have just tried to extend it... Should I bring it to SIG product?

Nope, this makes sense. This PR is totally reasonable, I was just making sure I hadn't missed such a categorization, if one had happened

All good :+1:

Rotfuks (approved):
fine for me

PR #2522, opened 2025-03-13
Fix: Set layout to single for overview introduction page

marians:
The page has file name `_index.md`, so Hugo assumes that it's a list page and applies the list template.

The PR adds the explicit layout setting.

Before

image

After

image


lyind (approved):
Looks good. `layout: single` is documented https://gohugo.io/templates/lookup-order/

PR #2529, opened 2025-03-14
Move MinAlertLevel

marians:
This should fix the config. It might produce a bigger result than we expect.


pipo02mix:
have you run `vale .` to see the output?

marians:
> have you run `vale .` to see the output?

Yes

pipo02mix (approved):

PR #2530, opened 2025-03-14
Fix duplicate H1 headline

marians:
This PR removes the double headline.

Before:

image


pipo02mix (approved):
good catch, what about include GPU as valid acronym?

PR #2555, opened 2025-04-03
Add margin to search box

pipo02mix:
Add the bottom margin as suggested by @marians


marians:
Tested this locally and with the container image built in CI, but can't see any effect.

image

image

pipo02mix:
you are right, the margin was not in the right context
Screenshot 2025-04-04 at 08 48 34

marians (approved):

PR #2595, opened 2025-05-20
Add default GS configuration

pipo02mix:

From support, we got a question about the default configuration of Zot and why custom images/registries were not used to cache the container images. I added the explanation

### DIFF
--- a/src/content/tutorials/registry/zot/index.md
+++ b/src/content/tutorials/registry/zot/index.md
@@ -10,7 +10,7 @@ menu:
 user_questions:
   - How can I cache container images within the cluster?
   - How can I have a backup registry for container images?
-last_review_date: 2024-10-31
+last_review_date: 2025-05-20
 owner:
   - https://github.com/orgs/giantswarm/teams/team-honeybadger
 ---
@@ -24,7 +24,7 @@ A registry cache within the cluster can provide several benefits.
 Here we explain how to set up a registry, using the [Zot](https://zotregistry.dev/) app provided by Giant Swarm. Zot is an OCI-native container image registry. The [Giant Swarm packaged version](https://github.com/giantswarm/zot) extends it with opinionated components like
 autoscaling, monitoring, Cilium network policies, etc.
 
-We explain how to deploy apps in our [app platform docs]({{< relref "/tutorials/fleet-management/app-platform/deploy-app/#creating-an-app-resource" >}}).
+In our platform, only the management cluster is created with default Zot installation for Giant Swarm images. When you want to have same functionality within your workload clusters, you need deploy the Zot app. We explain how to deploy apps in our [app platform docs]({{< relref "/tutorials/fleet-management/app-platform/deploy-app/#creating-an-app-resource" >}}).
 
 ## Zot configuration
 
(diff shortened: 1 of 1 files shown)


src/content/tutorials/registry/zot/index.md
```diff
@@ -24,7 +24,7 @@ A registry cache within the cluster can provide several benefits.
 Here we explain how to set up a registry, using the [Zot](https://zotregistry.dev/) app provided by Giant Swarm. Zot is an OCI-native container image registry. The [Giant Swarm packaged version](https://github.com/giantswarm/zot) extends it with opinionated components like
 autoscaling, monitoring, Cilium network policies, etc.
 
-We explain how to deploy apps in our [app platform docs]({{< relref "/tutorials/fleet-management/app-platform/deploy-app/#creating-an-app-resource" >}}).
+In our platform, only the management cluster is created with default Zot installation for Giant Swarm images. When you want to have same functionality within your workload clusters, you need deploy the Zot app. We explain how to deploy apps in our [app platform docs]({{< relref "/tutorials/fleet-management/app-platform/deploy-app/#creating-an-app-resource" >}}).
```
pipo02mix, line 27: @uvegla we dont cache WC container images, right?

uvegla: It depends. The MC Zot is configured with on-demand caching for gsoci: https://github.com/giantswarm/management-cluster-bases/blob/d3690fdea16362534cce08a887329f2bf4644894/extras/zot-cache/values.yaml#L140-L154. It is also exposed on ingress so its possible for the WCs to connect to the MC Zot for caching images. Moreover in the latest cluster chart MC zot is enabled by default: https://github.com/giantswarm/cluster/blob/v2.3.0/helm/cluster/values.yaml#L45-L48

So technically in that sense we do cache the gsoci images requested in WCs in the MC Zot instances.

The WCs do not get a Zot by default, they need to be manually added if needed. The benefit of that is being closer to the cluster and also customers can use it to cache their own images as well from private registries even. Technically customers could have their custom Zot on MCs too with their own caching rules and share that across their WCs. It depends on the use case.

pipo02mix:
@uvegla corrected PTAL

pipo02mix:
Finally, I explained how we cache images in GS and provided an entire example of configuration at the end, to serve as a working example for a customer. PTAL

src/content/tutorials/registry/zot/index.md
vxav: I suggested a couple minor changes but LGTM otherwise

src/content/tutorials/registry/zot/index.md
```diff
@@ -24,7 +24,9 @@ A registry cache within the cluster can provide several benefits.
 Here we explain how to set up a registry, using the [Zot](https://zotregistry.dev/) app provided by Giant Swarm. Zot is an OCI-native container image registry. The [Giant Swarm packaged version](https://github.com/giantswarm/zot) extends it with opinionated components like
 autoscaling, monitoring, Cilium network policies, etc.
 
-We explain how to deploy apps in our [app platform docs]({{< relref "/tutorials/fleet-management/app-platform/deploy-app/#creating-an-app-resource" >}}).
+The Giant Swarm Management cluster includes a [default Zot installation](https://github.com/giantswarm/management-cluster-bases/blob/main/extras/zot-cache/values.yaml#L52-L171) for managed component images. The cache comes configured as `on-demand`, so it only stores images when they're requested. The Zot instance running in the management cluster includes an ingress so the workload clusters can use it as a registry mirror too (for Giant Swarm images).
```
pipo02mix, line 27:
```suggestion
The Giant Swarm Management cluster includes a [default Zot installation](https://github.com/giantswarm/management-cluster-bases/blob/main/extras/zot-cache/values.yaml#L52-L171) for managed component images. The cache comes configured as `on-demand`, so it only stores images when they're requested. The Zot instance running in the management cluster includes an ingress so the workload clusters can use it as a registry mirror too (for Giant Swarm images).
```

PR #2747, opened 2025-07-11
Kernel settings

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/32687

### DIFF
--- a/src/content/overview/security/kernel-settings/index.md
+++ b/src/content/overview/security/kernel-settings/index.md
@@ -24,20 +24,23 @@ We adjust some kernel settings of Flatcar Container Linux machines used as Kuber
 
 ## General performance and security settings
 
-| Setting                                  | Value | Description                                                         |
-| :--------------------------------------- | :---- | :------------------------------------------------------------------ |
-| `kernel.kptr_restrict`                   | 2     | Hide kernel pointers to mitigate the kernel's attack surface.       |
-| `kernel.sysrq`                           | 0     | Reduce the kernel's attack surface.                                 |
-| `net.ipv4.conf.all.arp_ignore`           | 1     | Harden SSH security.                                                |
-| `net.ipv4.conf.all.arp_announce`         | 2     | Harden SSH security.                                                |
-| `net.ipv4.conf.all.log_martians`         | 1     | Log all martian packets coming to existing network interfaces.      |
-| `net.ipv4.conf.all.rp_filter`            | 1     | Harden SSH security.                                                |
-| `net.ipv4.conf.all.send_redirects`       | 0     | Don't send redirects for IPv4.                                      |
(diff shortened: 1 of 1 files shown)


marians:
You are adding the folder `mcp-docs` here, but it seems to have only empty files. Is that intentional?

glitchcrab (approved):
I don't see any issues with those values. we do configure some more on current flatcar instances, I pulled them from a running machine:

```
oa92854gef7a87 ~ # cat /etc/sysctl.d/99-sysctl.conf
net.bridge.bridge-nf-call-iptables=1
net.bridge.bridge-nf-call-ip6tables=1
net.ipv4.ip_forward=1
net.ipv6.conf.all.forwarding=1
net.ipv6.conf.all.disable_ipv6=0
net.ipv4.tcp_congestion_control=bbr
vm.overcommit_memory=1
kernel.panic=10
kernel.panic_on_oops=1
fs.inotify.max_user_instances=8192
fs.inotify.max_user_watches=524288
```

```
oa92854gef7a87 ~ # cat /etc/sysctl.d/hardening.conf
fs.inotify.max_user_watches = 16384
fs.inotify.max_user_instances = 8192
kernel.kptr_restrict = 2
kernel.sysrq = 0
net.ipv4.conf.all.log_martians = 1
net.ipv4.conf.all.send_redirects = 0
net.ipv4.conf.default.accept_redirects = 0
net.ipv4.conf.default.log_martians = 1
net.ipv4.tcp_timestamps = 0
net.ipv6.conf.all.accept_redirects = 0
net.ipv6.conf.default.accept_redirects = 0
# Increased mmapfs because some applications, like ES, need higher limit to store data properly
vm.max_map_count = 262144
# Reserved to avoid conflicts with kube-apiserver, which allocates within this range
net.ipv4.ip_local_reserved_ports=30000-32767
net.ipv4.conf.all.rp_filter = 1
net.ipv4.conf.all.arp_ignore = 1
net.ipv4.conf.all.arp_announce = 2

# These are required for the kubelet '--protect-kernel-defaults' flag
# See https://github.com/giantswarm/giantswarm/issues/13587
vm.overcommit_memory=1
kernel.panic=10
kernel.panic_on_oops=1
```

```
oa92854gef7a87 ~ # cat /etc/sysctl.d/99-zzz-override_cilium.conf

# Disable rp_filter on Cilium interfaces since it may cause mangled packets to be dropped
-net.ipv4.conf.lxc*.rp_filter = 0
-net.ipv4.conf.cilium_*.rp_filter = 0
# The kernel uses max(conf.all, conf.{dev}) as its value, so we need to set .all. to 0 as well.
# Otherwise it will overrule the device specific settings.
net.ipv4.conf.all.rp_filter = 0
```

PR #2833, opened 2025-09-29
Validate frontmatter via frontmatter-validator

marians:

Changes CI and Makefile to use https://github.com/giantswarm/frontmatter-validator for validating frontmatter, so that we have little duplication between docs and intranet.


lyind (approved):
LGTM

PR #2835, opened 2025-09-30
Use frontmatter-validator 0.2.0 with configuration system

marians:
frontmatter-validator 0.2.0 is more flexible, to be usable in docs and intranet.

This PR upgrades and adds the configuration for the two validation modes we need.


pipo02mix (approved):
nice

lyind (approved):
Very nice. I was thinking about such directory specific rules before but rejected the notion due to added complexity.

However, it looks like it should work.

PR #3178, opened 2026-06-17
feat: improving agent skills for style refinement

marians:
This PR

- Adds a new skill to find articles that need style improvements. It focuses on things that can be detected programmatically, like overly long text paragraphs, or headlines not using sentence case.
- Fixes a few skill details
- Improves the `improve-style` skill to fix more aspects, like wrapping too long shell commands.
- Fixes a CSS issue to enable line wraps for inline code spans


pipo02mix (approved):
nice

PR #3208, opened 2026-06-23
docs: restructure observability configuration section

marians:

First step in cleaning up section list pages (`_index.md`) that carry body content, which violates the convention in `AGENTS.md` that list pages should be frontmatter-only.

This PR restructures `/overview/observability/configuration/`:

- **Moves the configuration overview content** out of the `_index.md` list page into a dedicated `introduction/` page (`linkTitle: Introduction`, sorted first). The `_index.md` is now frontmatter-only.
- **Flattens the Grafana organization page** up one level, from `configuration/multi-tenancy/creating-grafana-organization/` to `configuration/creating-grafana-organization/`. An alias preserves the old URL.
- Because `multi-tenancy/` lost its only child, it's no longer a section — its `_index.md` becomes a leaf `index.md` (URL unchanged), so its body content no longer violates the list-page rule.
- Renames the subheading **"Creating a Grafana organization" → "Creating a GrafanaOrganization resource"**.
- Updates all 10 internal links repo-wide to the moved page.

**Resulting URLs**

- `/overview/observability/configuration/introduction/`
- `/overview/observability/configuration/multi-tenancy/`
- `/overview/observability/configuration/creating-grafana-organization/` (old URL aliased)

Verified with a local `hugo` build — no broken-ref warnings.

- [x] Internal links updated and build passes
- [ ] `last_review_date` left as-is since the prose itself is unchanged (only relocated)


src/content/overview/observability/configuration/introduction/index.md
```diff
@@ -0,0 +1,166 @@
+---
```
TheoBrigitte, line 1: This is a duplicate from the introduction/index.md I think

src/content/overview/observability/configuration/_index.md
```diff
@@ -9,157 +9,4 @@ menu:
 last_review_date: 2025-10-20
 owner:
   - https://github.com/orgs/giantswarm/teams/team-atlas
-user_questions:
```
TheoBrigitte, line 12: Please keep an alias to this location, so we do not end up with broken links.

src/content/overview/observability/configuration/introduction/index.md
marians: This _is_ `introduction/index.md`. The content is being moved from `configuration/_index.md`, as index pages are not supposed to deliver content, they're made for navigation.

src/content/overview/observability/configuration/_index.md
marians: This page is not being removed, so we don't need an alias.

TheoBrigitte: I mean this content is being moved to the `introduction/index.md`, I would prefer to have an alias to the new location, I assume that the current url is ending with `/observability/configuration` and the new one should be `/observability/configuration/introduction`.

src/content/overview/observability/configuration/introduction/index.md
TheoBrigitte: Yes my bad, sorry

src/content/overview/observability/configuration/_index.md
marians: We can't have an alias, as the old URL still exist. It's a valid navigation page.

TheoBrigitte: Ok, I am wondering where would the user land when they use a link to this page. Will they be able to find the actual introduction page ? Will this page have an index of the sub pages ?

marians: The page `/overview/observability/configuration/` will be an index page (list of child pages), where the first item in the list is the "Introduction" article.

TheoBrigitte: Ok that's fine with me then.

TheoBrigitte (approved):

PR #3220, opened 2026-06-25
feat: tag all pages with diataxis_content_type

marians:

Makes our docs Diátaxis-aware: every product article carries a `diataxis_content_type` frontmatter field (the [Diátaxis](https://diataxis.fr/) quadrant — `tutorial`, `how-to-guide`, `reference`, or `explanation`), and CI enforces it.

**Backfill**
- **203 articles tagged** across `overview/`, `getting-started/`, `tutorials/`, `reference/`, from a full assessment of the corpus (each page read and classified, mixed/borderline pages double-checked).
- The CRD and Helm-chart reference **generators** (`crd.template`, `chart.template`) now emit `diataxis_content_type: reference`, so regenerated pages keep the field.

**Enforcement** (now that `frontmatter-validator` v0.7.0 is released)
- Bumped `frontmatter-validator` to **v0.7.0** in both the pre-commit hook and the CI install step (`validate.yaml`).
- Enabled `NO_DIATAXIS_CONTENT_TYPE` (required on articles, skipped for `_index.md` list pages) and `INVALID_DIATAXIS_CONTENT_TYPE` (value must be a valid quadrant) in `config-default.yaml`.
- Relaxed `NO_DIATAXIS_CONTENT_TYPE` for sections outside the Diátaxis scope (`changes`, `meta`, `search`, `support`, `vintage`) and for `.claude` skill files. `INVALID_DIATAXIS_CONTENT_TYPE` stays on everywhere (it only fires on a bad value).

### Notes for reviewers

- Verified with the v0.7.0 binary across the **entire** `src/content`: **zero Diátaxis violations**.
- This PR touches 203 pages, so the front-matter validator surfaces ~52 **pre-existing** `REVIEW_TOO_LONG_AGO` annotations (stale review dates) on some of them. These are **not** caused by this change and are not Diátaxis issues — `last_review_date` was intentionally **not** bumped (mechanical change, no content review).
- Minor follow-up: local `make validate-front-matter` still pins an older image (`0.3.2`); CI uses the v0.7.0 binary via `install-binary-action`. Bump the Makefile image once a 0.7.0 image is published.
- Per-page type assignments (with rationale and mixed-page flags) are available for reviewers who want to spot-check classifications.


pipo02mix (approved):
LGTM I did not check all but the ones in tutorial and overview looks legit

PR #3231, opened 2026-06-26
docs: add HelmRelease pointers across pages with passing App-CR mentions

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/36892

Correct reference pages that point yet to App CR. Each edit adds a short note pointing readers at the HelmRelease alternative or the App management overview, without rewriting the existing content.


mproffitt (approved):
Docs review: **Approve**.

Reviewed as part of the App Platform → Flux HelmRelease documentation migration, validated locally against an integrated build of the full batch. Migration framing and technical content are sound; remaining items (Vale spelling, minor clarity nits) are non-blocking follow-ups.

PR #3284, opened 2026-07-16
Add AI agents authentication deep dive

fiunchinho:
Adds a new explanation page under **Overview → AI agents**: a token-by-token walkthrough of how user identity flows from an MCP client through Muster and mcp-kubernetes to the Kubernetes API.

Covers: the login redirect chain (CIMD, PKCE, Dex, upstream IdP), Muster's per-request token validation, what each token actually contains (annotated examples), the `forward` and `exchange` downstream modes, impersonation for workload clusters, direct Dex tokens for non-interactive clients, outbound OAuth for backends with a different identity provider (github-mcp), and how in-cluster agents act on behalf of a human.

All content fact-checked against `muster`, `mcp-oauth`, `mcp-kubernetes`, and `klaus-gateway` on `main` as of 2026-07-16 (post muster#947/#968: Muster issues only opaque tokens and never signs JWTs; Dex is the sole SSO authority).

Supersedes the intranet version in giantswarm/giantswarm#37117, which will be closed in favor of this public page.

Complements the existing [Security](https://docs.giantswarm.io/overview/ai-agents/security/) and [Architecture](https://docs.giantswarm.io/overview/ai-agents/architecture/) pages — this one is the deep dive; those stay the quick answers.


src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,716 @@
...
+
+### 3. Does Dex vouch for the user right now
+
+On every request coming from MCP clients, Muster will call the `/userinfo` endpoint in Dex to check if the Dex access token is still valid at the Dex level. This catches revocation at Dex immediately (session deleted or revoked at Dex). Note that this check only proves the session is still alive at Dex. A user removed at the upstream identity provider (Google / Azure) isn't visible to `/userinfo`. Dex itself only finds out the next time it refreshes against the upstream, which is [check 2 described earlier](#2-is-the-underlying-dex-session-still-alive). Each layer vouches for itself in real time. News from the upstream layer arrives at refresh time.
+
+### 4. Was this token minted for this server
+
+When it mints a token, Muster records in its storage which resource the token was requested for (its own resource URI—this is OAuth's "resource indicator," RFC 8707). At validation time it checks that binding, so a token minted for a different resource server is rejected even if it's otherwise valid.
```
marians: For such step-by-step content we have the HUGO shortcode `steps` and `step`. Would be nice to use it here. You'll find examples in the code base.

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,716 @@
...
+
+![Claude Code's /mcp view showing MCP servers that need authentication](claude-mcp-needs-auth.png)
+
+Muster is itself the Authorization Server for MCP clients like Claude Code. Behind the scenes, it delegates the actual user authentication to Dex. Muster is registered as a client of Dex, which lets us plug in different identity providers (Dex connectors) like Google or Azure without changing the architecture.
+
+## What happens when you start the OAuth flow to log in to Muster
+
+### 1. Claude Code → Muster
+
+When Claude Code gets a `401 Unauthorized` HTTP response, the response carries a `WWW-Authenticate` header pointing at the server's metadata, which is how Claude Code discovers the authorization server and its endpoints. Using that information, Claude Code (the client) opens your browser at Muster's `/oauth/authorize`, identifying itself with its `client_id`. For MCP clients that `client_id` isn't a pre-registered value but a URL the client controls (see ["How Claude Code becomes a client without pre-registration"](#how-claude-code-becomes-a-client-without-pre-registration) below).
```
pipo02mix:
```suggestion
When Claude Code gets a `401 Unauthorized` HTTP response, the response carries a `WWW-Authenticate` header pointing at the server's metadata, which is how Claude Code discovers the Authorization server (AS) and its endpoints. Using that information, Claude Code (the client) opens your browser at Muster's `/oauth/authorize`, identifying itself with its `client_id`. For MCP clients, that `client_id` isn't a pre-registered value but a URL the client controls (see ["How Claude Code becomes a client without pre-registration"](#how-claude-code-becomes-a-client-without-pre-registration) below).
```

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,716 @@
...
+
+Dex is doing the exact same trick one level deeper: it's an AS to Muster but a client of the identity provider, for example Google. Each Dex "connector" holds a client_id/secret registered with the upstream identity provider. It redirects the browser to the identity provider's login page.
+
+### 4. User logs in
+
+You type your password at the identity provider login page, and only there. Neither Dex, nor Muster, nor Claude Code ever sees it.
+
+### 5. The chain unwinds
+
+The identity provider redirects the browser back to Dex's callback with a code. Dex exchanges it (authenticating with its identity provider / connector `client_secret`), mints its own tokens, and redirects the browser to Muster's `/oauth/callback` with a new code. Muster exchanges that code at Dex's `/token` endpoint, authenticating with Muster's Dex `client_secret`, and receives the Dex ID token. Muster then redirects the browser one last time to Claude Code's localhost callback with yet another code. When Claude Code exchanges that code, Muster mints and returns its own access token.
```
pipo02mix, line 67: IMO it is a bit confusing, explaning why exchange the cods may help users to understand why so many tokens are involved

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,716 @@
+---
+title: 'Authentication deep dive: from your MCP client to the Kubernetes API'
+linkTitle: Authentication deep dive
```
marians: I'd prefer a shorter handle in the menu here

Image

```suggestion
linkTitle: Authentication
```

src/content/overview/ai-agents/authentication/index.md
pipo02mix:
```suggestion
When Dex mints a token, Muster records in its storage which resource the token was requested for (its own resource URI—this is OAuth's "resource indicator," RFC 8707). At validation time, it checks that binding, so a token minted for a different resource server is rejected even if it's otherwise valid.
```

src/content/overview/ai-agents/authentication/index.md
marians: This also looks like a good case for step/steps (see my other comment)

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,716 @@
...
+
+Can any client just request tokens addressed to any other client? No, this is Dex's cross-client audience feature, and it's deny-by-default. The target client's Dex config must explicitly list the requester in its `trustedPeers` config.
+
+Similarly, `mcp-kubernetes` is configured to accept tokens that contain `muster` in the audience field, via the `trustedAudiences` list in `mcp-kubernetes` configuration. This is empty by default (deny-by-default: each aggregator is an explicit entry).
+
+The same audience discipline appears one more time, in the token exchange scenario. When Muster exchanges tokens with a target Dex, the target Dex is itself a validator of an incoming ID token—one whose audience is `muster`, not the target Dex. It accepts it because its connector is explicitly configured with which audience to expect on incoming tokens: Muster's client id at the source Dex. A Dex instance has no audience of its own as an issuer. Like every other validator in this document, it accepts exactly the audiences its configuration names, and nothing else.
```
pipo02mix: Does it mean all Dex instances are configured with the audience of the central Dex?

src/content/overview/ai-agents/authentication/index.md
QuentinBisson: Should the title not be agent platform instead or AI Agent?

QuentinBisson: it makes me think fleet management is now some kind of misdirection right? It is a fleet of what ?

src/content/overview/ai-agents/authentication/index.md
fiunchinho: done, ptal

src/content/overview/ai-agents/authentication/index.md
fiunchinho: I changed the handle in the menu. 

> Should the title not be agent platform instead or AI Agent?

Renaming it would touch every page in the section plus its menu identifier. I'd rather not couple that to this page; happy to discuss it in a separate issue/PR

src/content/overview/ai-agents/authentication/index.md
fiunchinho: done

src/content/overview/ai-agents/authentication/index.md
fiunchinho: Actually, this check is about Muster's own opaque access token, which Muster mints and stores (Dex never sees it). I replaced the pronoun for the explicit name.

src/content/overview/ai-agents/authentication/index.md
fiunchinho: In practice all MCs, yeah. Each MC runs a connector (`giantswarm-simple-oidc`) configured to trust the central Dex as issuer and to expect Muster's client ID as the audience on incoming tokens. It's provisioned per management cluster together with the token-exchange credentials.

src/content/overview/ai-agents/authentication/index.md
fiunchinho: Added a small paragraph trying to explain it better.

src/content/overview/ai-agents/authentication/index.md
QuentinBisson: Oh you're totally right. I think this is more of a question to @sig-docs

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,744 @@
...
+  CC->>M: exchange code + PKCE verifier
+  M-->>CC: Muster access token (opaque)
+  Note over M: stores map:<br/>Muster token → Dex token set
+  Note over CC: Claude Code authenticates to Muster<br/>with the opaque token only
+{{< /mermaid >}}
+<!-- vale on -->
+
+### OAuth clients
+
+In OAuth, the client is the application that wants tokens, and the user never types a password into it. A client redirects the user's browser to the Authorization Server (AS) and waits for the browser to come back with a one-time authorization code. It then exchanges that code for tokens over a back-channel call, identifying itself with its own credentials (`client_id` + `client_secret` for confidential clients like Muster). Public clients like Claude Code have no secret and use PKCE instead—see [below](#how-claude-code-becomes-a-client-without-pre-registration). Any application can be a client in one relationship and an Authorization Server in another. For example, Claude Code is a client of Muster, which is an AS to Claude Code but a client of Dex.
```
pipo02mix:
```suggestion
In OAuth terms, the client is the application that requests tokens, and the user never types a password. A client redirects the user's browser to the Authorization Server (AS) and waits for the browser to return a one-time authorization code. It then exchanges that code for tokens over a back-channel call, identifying itself with its own credentials (`client_id` + `client_secret` for confidential clients like Muster). Public clients like Claude Code have no secret and use PKCE instead—see [below](#how-claude-code-becomes-a-client-without-pre-registration). Any application can be a client in one relationship and an AS in another. For example, Claude Code is a client of Muster, which is an AS to Claude Code but a client of Dex.
```

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,744 @@
...
+
+So MCP clients identify themselves using Client ID Metadata Documents (CIMD). The `client_id` is an HTTPS URL, controlled by the client's vendor, hosting a JSON document that describes the client (its name, its redirect URIs). When Muster receives an authorization request with a URL-shaped `client_id`, it fetches that document and treats it as the client's registration. Nothing is stored ahead of time, and no secrets are exchanged between Muster's team and the client's vendor. This has two useful properties. First, client identity is anchored to domain ownership: only Anthropic can serve a document under an anthropic.com URL. So a malicious app can't credibly present itself as Claude Code. Second, every installation of the same client shares the one `client_id` (the URL), so Muster keeps no per-laptop registration state.
+
+(Muster also supports the older mechanism, Dynamic Client Registration (RFC 7591), as a fallback for clients that don't publish a metadata URL. The client POSTs its description to `/oauth/register` and receives a server-generated random `client_id`.)
```
pipo02mix:
```suggestion
**Note**: Muster also supports the older mechanism, Dynamic Client Registration (RFC 7591), as a fallback for clients that don't publish a metadata URL. The client POSTs its description to `/oauth/register` and receives a server-generated random `client_id`.
```
We dont use parethesis in rest of docs

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,744 @@
...
+| Dex access token | Dex | The user's Dex session | Dex itself (spent at `/userinfo`) | Muster, server-side | Dex, on every `/userinfo` call (check 3) |
+| Dex refresh token | Dex | The user's Dex session | Dex itself | Muster, server-side | Dex, on refresh (check 2)—this is where upstream deprovisioning surfaces |
+
+## What do these tokens actually look like
+
+Muster's access token is opaque: a random string with no internal structure.
+
+**Muster access token—opaque. What Claude Code sends on every request:**
+
+```text
```
pipo02mix:
```suggestion
Muster's access token is opaque: a random string with no internal structure. Look an example of what Claude Code sends on every request:

```text
```

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,744 @@
...
+    "access_token": "...",
+    "refresh_token": "..."
+  }
+}
+```
+
+The Dex ID token is different: it's a JWT, three base64url-encoded segments separated by dots (header, payload, signature).
+
+**Dex ID token—a signed JWT. Raw form, then the first two segments decoded (annotated):**
+
```
pipo02mix:
```suggestion
The Dex ID token is different: it's a JWT, three base64url-encoded segments separated by dots (header, payload, signature). Look at an example of an ID token. Raw form, then the first two segments decoded (annotated):

```

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,744 @@
...
+  "email": "user@giantswarm.io",               // who it is about
+  "groups": ["giantswarm:admins"],             // what kubernetes RBAC will evaluate
+  "iat": 1751636400,                           // minted at
+  "exp": 1751638200                            // expires at           (Expiration check)
+}
+```
+
+Two things worth noticing. First, base64url is an encoding, not encryption: anyone holding this token can read every claim. The signature (the third segment) prevents *modification*, not reading. Second, each claim maps to one of the checks in ["How are Dex's ID tokens validated"](#how-are-dexs-id-tokens-validated) below. `iss`, `aud` and `exp` are the envelope. `email` and `groups` are the identity that will reach the Kubernetes RBAC evaluation and audit log.
+
+**Muster refresh token—opaque. What Claude Code redeems when the access token expires:**
```
pipo02mix:
```suggestion
Now, look what Claude Code redeems when the access token expires:
```

src/content/overview/ai-agents/authentication/index.md
fiunchinho: Good catch on the ambiguous pronoun — but I went the other direction: this check is about Muster's own opaque access token, which Muster mints and stores (Dex never sees it). The RFC 8707 resource binding lives in Muster's token storage, which is why the check works at all. I've reworded it to "When Muster mints an access token, it records in its storage…" so the subject is explicit. Thanks for flagging it!

src/content/overview/ai-agents/authentication/index.md
fiunchinho: @QuentinBisson that's a fair question, but it's about the parent section (`overview/ai-agents`), which predates this PR — renaming it would touch every page in the section plus its menu identifier. I'd rather not couple that to this page; happy to discuss it in a separate issue/PR if you want to push for "Agent platform".

src/content/overview/ai-agents/authentication/index.md
fiunchinho: Applied, with one small keep: I left "the user never types a password **into it**" intact. Dropping "into it" would make the sentence say the user never types a password at all — but they do, at the upstream identity provider; the point is that the *client* never sees it.

src/content/overview/ai-agents/authentication/index.md
fiunchinho: Applied to all three token examples (and the storage-record card too, for consistency) — the bold pseudo-headers are gone in favor of prose lead-ins. I adjusted the wording slightly ("Look **at** an example…", "look **at** what Claude Code redeems…") while keeping the structure you proposed.

.vale/styles/config/vocabularies/docs/accept.txt
```diff
@@ -25,6 +26,7 @@ Autoscaler
 [Ii][Aa][Mm]
 [Kk]ustomization[s]?
 [Kk]ustomize
+lookups
```
QuentinBisson:
```suggestion
lookup[s]?
```

.vale/styles/config/vocabularies/docs/accept.txt
```diff
@@ -168,6 +170,7 @@ randomly
 rapidly
 RBAC
 really
+reconnections
```
QuentinBisson:
```suggestion
reconnection[s]?
```

.vale/styles/config/vocabularies/docs/accept.txt
```diff
@@ -198,6 +201,8 @@ TraceQL
 UDP
 unmanaged
 upstream[s]?
+URIs
+validators?
```
QuentinBisson:
```suggestion
validator[s]?
```

src/content/overview/ai-agents/authentication/index.md
```diff
@@ -0,0 +1,738 @@
...
+last_review_date: 2026-07-16
+user_questions:
+  - Which token does my MCP client send to Muster?
+  - How does Muster validate tokens on every request?
+  - How does my identity reach the Kubernetes API and its audit log?
+  - How does Muster reach clusters that use a different Dex instance?
+  - How do AI agents act on behalf of a human?
+---
+
+You can configure several different MCP servers in Claude Code (or Cursor, and other MCP clients), and these MCP servers may or may not require authentication. For example, you may want to investigate the state of a Kubernetes cluster using `mcp-kubernetes`, which requires Kubernetes authentication. The challenge here is to keep the user identity through the whole communication, so at the very end, the requests to the Kubernetes API carry the user identity (and permissions).
```
QuentinBisson, line 23: What do you think about adding an introduction here that 1 points to the what muster is and oauth docs with links I think and 2 explain what muster authentication mechanisms are ? Becasue it's not really saying it supports oauth (token forwarding, token exchange or connecting to a remote authorization server right? I feel like this might simplify the content below more and make for a more shared understanding? Then for each flow we could add some diagrams to explain it best?

QuentinBisson: I'm not reviewing the content here but the format. What do you think about adding the main page to be the introduction and then have subpages like this:
- 1 page one how Muster authentication and oauth work with MCPs that might be a bit more high level
- another page that deep dives into how claude code access muster and the MCps behind it if necessary
- page on agent access to muster and OBO?

That way we can also introduce the agentgateway in the first one?

QuentinBisson: I'm sorry if this is a bit late in the process but the discussion frm yesterday shed some lights and I don't think we could have a shared understanding without splitting this into digestable chunks and without diagrams

fiunchinho: We do have diagrams. Please review using the local server to render the page properly and see all the content the way it's intented to be shown.

fiunchinho: I believe splitting the content in several smaller pages makes things actually harder to follow. In this page, we are telling the story about what happens when the user types `/mcp` in their claude code session. Splitting it in several pages makes it uncomfortable to follow and harder to consume IMO.

Also, this style was chosen purposely: instead of throwing definitions around (what's muster, what's oauth, etc), we explain why is necessary and what's happening.

marians (approved):

pipo02mix (approved):

PR #3352, opened 2026-08-19
docs: reclassify GitOps tooling as how-to, extract a reference page

weatherhog:

"Tools for your GitOps workflow" declared `diataxis_content_type: reference` but is troubleshooting guidance — it tells you which tool to reach for when Flux misbehaves. The lookup material (command, behavior, source per tool) belongs on a reference page.

- The tools page becomes a `how-to-guide` and gains the up-front summary list it was missing. Today `test-all-ff` and `validate.sh` are only named mid-section, so you can't see the toolbox before reading the whole page.
- **New reference page** at `/reference/gitops-tooling/`: every tool with its command, what it does, and where it comes from, split into local build/validation and cluster inspection. `reference/kubectl-gs/gitops/` already exists but covers only kubectl-gs subcommands, not `fake-flux` / `test-all-ff` / `validate.sh`.
- The two pages cross-link with `relref`.

- `last_review_date` bumped: the page sat at `2025-03-14`, past the validator's 365-day `REVIEW_TOO_LONG_AGO` gate, and this is a real content revision.

### DIFF
--- a/src/content/reference/gitops-tooling/index.md
+++ b/src/content/reference/gitops-tooling/index.md
@@ -1,0 +1,41 @@
+---
+linkTitle: GitOps tooling
+title: GitOps tooling reference
+diataxis_content_type: reference
+description: The commands, sources, and behavior of the tools used to build and validate Flux manifests locally, in one lookup table.
+weight: 50
+menu:
+  principal:
+    parent: reference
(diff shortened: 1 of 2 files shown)


pipo02mix (approved):
nice

PR #3353, opened 2026-08-19
docs: move cluster config layering out of the base template how-to

weatherhog:

"Creating a base template for your workload cluster" is correctly typed as a `how-to-guide`, but it opened with a block of pure explanation. That material moves to the cluster configuration concept page, which already covers the three cluster config sources and is the natural home for it.

**On `cluster-configuration` (team-phoenix-owned):**

- Gains the `cluster-<provider>` → `cluster` → `cluster-shared` chart dependency chain.
- Gains the `config` / `extraConfigs` merge behavior and the merging-configmaps RFC link.
- Gains the `linkTitle` it was missing. The frontmatter validator requires it alongside a `menu` block, so the page would otherwise fail `validate` as soon as anyone touches it.

**On the bases page:** the intro becomes a two-sentence lead-in plus a `relref`, and four bugs are fixed:

- The example `App` resource was missing `metadata:`, so `name` sat directly under `kind`.
- Step 1 creates `bases/clusters/capa/v0.21.0`, but step 4's two `sed` commands wrote to `.../0.21.0/` — so they silently did nothing.
- Step 3 called the file `cluster-config.yaml`; every command and every other step uses `cluster_config.yaml`.
- Missing space in "…yaml`.Extract the values".

The prose now says `extraConfigs`, which is the actual App CR field name (it said `extraConfig`).

- `last_review_date` bumped on both pages — real content changes on both sides.
- @giantswarm/team-phoenix you own `cluster-configuration` — the added sections are the ones that were duplicating your page from the bases guide.
- Noticed but deliberately **not** changed: step 6 of the bases guide still uses `patchesStrategicMerge`, which Kustomize has deprecated. Out of scope here; worth a follow-up.

### DIFF
--- a/src/content/overview/fleet-management/cluster-management/cluster-concepts/cluster-configuration/index.md
+++ b/src/content/overview/fleet-management/cluster-management/cluster-concepts/cluster-configuration/index.md
@@ -1,4 +1,5 @@
 ---
+linkTitle: Cluster configuration
 title: Cluster configuration
 diataxis_content_type: explanation
 description: Explanation of how cluster configuration works in Cluster API and how you can customize it.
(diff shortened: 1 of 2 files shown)


pipo02mix (approved):
LGTM

PR #3354, opened 2026-08-19
docs: clean up the chart metadata reference

weatherhog:

The chart metadata page is correctly typed `reference` and needs no split — it's hand-written (not one of the schema-generated CRD pages), so it just needed a cleanup pass:

- Removed the `TODO: Add information about how to align keyword usage…` note, which was visible in the published page.
- Fixed the typos "requriements" and "authered", and the garbled terminology entry "**Legacy Chart.yaml key** key used previously in…".
- Added `namespace-singleton` and `gpu-instances` to the complex example. Both are documented but appeared in neither example.
- **Bug:** the complex example had `io.giantswarm.application.restrictions.fixed-namespace: "false"`, but that field takes the namespace the app must be installed into (documented example: `helloworld`), not a boolean.
- Bolded the `OPTIONAL` marker so it matches `**REQUIRED**` and `**RECOMMENDED**` (10 occurrences).
- Trimmed the OCI-migration "Background" from three paragraphs to two and linked to the app management overview. Every fact is kept — the "legacy key" terminology and the CI/CD translation note are load-bearing for the rest of the page.
- **Added the missing `weight`.** The page has a `menu` block, and frontmatter-validator exits non-zero without a `weight`. Because CI pipes into the validator under `bash -eo pipefail`, any PR touching this page would have failed `validate` before this. `weight: 200` sorts it after `cluster-apps` (100) and before `annotations`/`labels` (500).

- `last_review_date` bumped — real content revision.


pipo02mix (approved):
Lgtm

PR #3355, opened 2026-08-19
docs: fix Diataxis types and bugs on the legacy App CR pages

weatherhog:

Part of giantswarm/giantswarm#37092 (sub-issue of giantswarm/giantswarm#37021) — the mechanical PR for the four pages documenting the legacy `App` custom resource.

The `App` CR is deprecated in favor of Flux HelmRelease, so per the issue these pages get frontmatter corrections and genuine bug fixes only. No content is moved and no pages are created: the reference material can't move into the schema-generated CRD page, and it isn't worth investing further in deprecated content.

- **App configuration** and **Defaulting and validation**: both declared `explanation` but describe machinery (field tables, merge algorithms, webhook behavior), so they're now `reference`.
- **App configuration**: six Kubernetes object manifests used `kind: configmap` / `kind: secret`, which Kubernetes rejects. Now `ConfigMap` / `Secret`. Note the lowercase `configMap` / `secret` values under `spec.extraConfigs[].kind` are the App CR's *own field values* and are correctly left alone — a blanket replace here would have broken the docs.
- **App configuration**: removed a stray unbalanced backtick in the merge-order list ("P = 125, position in the list: \`1").
- **App bundle**: `[Hello World app]` was a reference-style link with no matching definition anywhere in the file, so it rendered as literal brackets. It now points at the repo. Type was already correct (`explanation`).
- **App Sets**: demoted the body `# Creating app sets` to an H2. The page title comes from frontmatter, so the body must not carry an H1. The `#` lines inside the fenced YAML blocks are comments and are untouched. Type was already correct (`how-to-guide`).

- `last_review_date` deliberately **not** bumped on any of these pages. Per giantswarm/giantswarm#37021, a `diataxis_content_type` correction is mechanical rather than a content review, and all four dates are well inside the 365-day window.
- Verified locally: frontmatter-validator v0.7.0 (the version CI installs) clean, markdownlint clean, Hugo build with no warnings. Vale reports pre-existing errors on these pages, but none on lines this PR touches — `validate-prose` runs reviewdog with `--filter-mode=added`.

### DIFF
--- a/src/content/tutorials/continuous-deployment/apps/app-sets.md
+++ b/src/content/tutorials/continuous-deployment/apps/app-sets.md
@@ -19,7 +19,7 @@ last_review_date: 2026-07-02
 
 This document is part of the documentation to use GitOps with Giant Swarm app platform. You can find more information about the [app platform in our docs]({{< relref "/overview/fleet-management/app-management/" >}}).
 
-# Creating app sets
+## Creating app sets
 
 It's often desirable to deploy a group of apps together, as a single deployment step. In Giant Swarm, it's called `App Sets`. There's nothing special about `App Sets`: they aren't a separate API entity, but rather just a configuration pattern enabled by [`Kustomize`](https://kustomize.io/) and [`Flux`](https://fluxcd.io/flux/). The purpose brings you the following benefits:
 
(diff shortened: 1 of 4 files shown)


pipo02mix (approved):
LGTM

